PR #22, opened 2024-07-22
breaking: uppercase duration representations

punkeel:
All the ISO8601 resources I could find use uppercased representations for Durations/Periods.

3.4.1 notes:
> NOTE 1: In date and time representations lower case characters may be used when upper case characters are not available.

3.4.3 notes:
> In representations of duration (4.4.3.2), the following designators are used as part of the expression: [Y] [M] [W] [D] [H] [M] [S]

This change introduces a breaking change: durations are now emitted fully uppercased.

It looks like other projects are lenient in what they accept - and strict in emitting uppercase.
Spring Boot:
https://github.com/spring-projects/spring-boot/blob/4d466c3cc097/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/convert/DurationStyleTests.java

Temporal:
https://github.com/tc39/test262/blob/880f8a5ba6/test/built-ins/Temporal/Duration

**I understand this is quite a highly pedantic change** :D

### DIFF
--- a/COMPARE.md
+++ b/COMPARE.md
@@ -288,7 +288,7 @@ use jiff::{Span, ToSpan};
 fn main() -> anyhow::Result<()> {
     let span = 5.years().months(2).days(1).hours(20);
     let json = serde_json::to_string_pretty(&span)?;
-    assert_eq!(json, "\"P5y2m1dT20h\"");
+    assert_eq!(json, "\"P5Y2M1DT20H\"");
 
     let got: Span = serde_json::from_str(&json)?;
     assert_eq!(got, span);
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -1711,11 +1711,11 @@ impl Date {
     ///
     /// // The default limits durations to using "days" as the biggest unit.
     /// let span = d2.since(d1)?;
-    /// assert_eq!(span.to_string(), "P8456d");
+    /// assert_eq!(span.to_string(), "P8456D");
     ///
     /// // But we can ask for units all the way up to years.
     /// let span = d2.since((Unit::Year, d1))?;
-    /// assert_eq!(span.to_string(), "P23y1m24d");
+    /// assert_eq!(span.to_string(), "P23Y1M24D");
     ///
     /// # Ok::<(), Box<dyn std::error::Error>>(())
     /// ```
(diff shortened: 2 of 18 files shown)


BurntSushi:
I don't think this would be a breaking change.

But the current behavior is very specifically intended, and it is consistent with the [Temporal ISO 8601 grammar](https://tc39.es/proposal-temporal/#sec-temporal-iso8601grammar) that Jiff is specifically documented to support.

I'm open to changing to uppercase by default here, but I'd like to better understand whether this is actually necessary. I used lowercase for the unit designators by default because I find it creates duration strings that are way easier to read. When everything is uppercase, then the numbers and letters all run together for me.

punkeel:
> I find it creates duration strings that are way easier to read

I agree with that.

I'm not aware of this being a _hard requirement_- and seeing as this was intended, I'm happy to close that PR 👍

falsetru:
Pydantic only accepts uppercased representations.
https://docs.pydantic.dev/latest/api/standard_library_types/#datetimetimedelta

```python
>>> import datetime
>>> import pydantic
>>> Delta = pydantic.TypeAdapter(datetime.timedelta)
>>> Delta.validate_json('"P23y1m24d"')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/falsetru/.virtualenvs/targetyo/lib/python3.12/site-packages/pydantic/type_adapter.py", line 142, in wrapped
    return func(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/falsetru/.virtualenvs/targetyo/lib/python3.12/site-packages/pydantic/type_adapter.py", line 391, in validate_json
    return self.validator.validate_json(data, strict=strict, context=context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for timedelta
  Input should be a valid duration, quantity invalid in date part of duration [type=time_delta_parsing, input_value='P23y1m24d', input_type=str]
    For further information visit https://errors.pydantic.dev/2.8/v/time_delta_parsing
>>> Delta.validate_json('"P23Y1M24D"')
datetime.timedelta(days=8449)
```

punkeel:
@falsetru worth opening an issue on Pydantic's side, regardless of what happens here?

@BurntSushi up to you, I guess - I'm happy to reopen/rebase/fix this PR if you change your mind

BurntSushi:
Note that I ended up making this change. See #190 for more details. This is now the default in `jiff 0.1.19` on crates.io.

punkeel:
Oh no, I missed my chance to contribute! 😛 😂

PR #23, opened 2024-07-22
Optimize common functions with Neri-Schneider

DaniPopes:
Optimize `common.rs` by using Neri-Schneider algorithms, as presented in CppNow 2023: <https://www.youtube.com/watch?v=0s9F4QWAl-E>

There are likely more places where they can be applied, but I just wanted to open this PR for now :)

### DIFF
--- a/src/util/common.rs
+++ b/src/util/common.rs
@@ -1,37 +1,48 @@
-/// A collection of datetime related utility functions.
-///
-/// # Constant functions
-///
-/// This module also contains some utilities for making some API items `const`.
-///
-/// One of the main reasons why more of this crate isn't `const` is because of
-/// our ranged integer types. We very specifically use those types wherever we
-/// do arithmetic as a mitigation for the absolute insance boundary conditions
-/// that one must deal with in a datetime library. Unfortunately, these range
-/// integer types are very difficult to make use of in `const` context. One of
-/// the main blockers is the fact that Rust doesn't support `const` methods in
-/// traits. That means we'd be unable to do `x + y` where `x` and `y` are range
-/// integers. (Normal `x + y` arithmetic only works in `const` today because of
-/// special support for Rust's standard integer types.)
-///
-/// With that said, it is sometimes useful to provide APIs that are const. For
-/// example, `Date::new`. But we still need to do some checking on the inputs
-/// to ensure they are valid. Since most of our arithmetic code is defined
-/// using range integers, we have to re-write some it here using standard
-/// primitive types.
-///
-/// In general, this code should not be used unless it is specifically required
-/// to make something `const` AND if it is worth making it `const` in the first
-/// place. Remember, much of this library _could_ be `const` if we were willing
-/// to write `const`-compatible code, but I deemed it far too annoying to do
-/// so.
-use crate::util::t::{Day, Month, Year, C};
+/*!
+A collection of datetime related utility functions.
+
+# Constant functions
+
+This module also contains some utilities for making some API items `const`.
+
+One of the main reasons why more of this crate isn't `const` is because of
+our ranged integer types. We very specifically use those types wherever we
+do arithmetic as a mitigation for the absolute insance boundary conditions
+that one must deal with in a datetime library. Unfortunately, these range
+integer types are very difficult to make use of in `const` context. One of
+the main blockers is the fact that Rust doesn't support `const` methods in
+traits. That means we'd be unable to do `x + y` where `x` and `y` are range
+integers. (Normal `x + y` arithmetic only works in `const` today because of
+special support for Rust's standard integer types.)
+
+With that said, it is sometimes useful to provide APIs that are const. For
+example, `Date::new`. But we still need to do some checking on the inputs
+to ensure they are valid. Since most of our arithmetic code is defined
+using range integers, we have to re-write some it here using standard
+primitive types.
+
+In general, this code should not be used unless it is specifically required
+to make something `const` AND if it is worth making it `const` in the first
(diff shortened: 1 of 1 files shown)


src/util/common.rs
```diff
@@ -60,54 +74,36 @@ pub(crate) fn saturate_day_in_month(
...
 pub(crate) fn days_in_month(year: Year, month: Month) -> Day {
-    if month == 2 && is_leap_year(year) {
-        Day::N::<29>()
-    } else {
-        days_in_month_common_year(month)
-    }
+    Day::new_unchecked(days_in_month_unranged_unchecked(
+        year.get(),
+        month.get() as u8,
+    ))
```
BurntSushi, line 82: You've gotta stay in ranged integer land here. As soon as you start creating new ranged integers like this, the constructor can't tell what the min/max values are allowed to be, and so it assumes the worst case.

There's a reason why there is duplicated code in this module. You can't use unranged code to implemented arithmetic for ranged integers, and you can't use ranged integers inside `const` functions.

DaniPopes: Fixed in https://github.com/BurntSushi/jiff/pull/23/commits/bfefca72b020141c7478a2f877a0d9eff1857410.

FYI just adding `year.get()` in `days_in_month` panics with
```
thread 'civil::iso_week_date::tests::prop_next_day_is_greater' panicked at src/util/rangeint.rs:2173:1:
ri16 max 10000 is not in range -9999..=9999
```

src/util/common.rs
```diff
@@ -60,54 +75,41 @@ pub(crate) fn saturate_day_in_month(
...
+    if month == 2 {
+        if is_leap_year(year) {
+            Day::N::<29>()
+        } else {
+            Day::N::<28>()
+        }
     } else {
-        days_in_month_common_year(month)
+        let month = month.get();
+        if (month ^ month >> 3) & 1 == 0 {
```
DaniPopes, line 88: Do you want me to implement bit operations for range int?

src/util/common.rs
```diff
@@ -1,53 +1,68 @@
...
 /// Returns true if and only if the given year is a leap year.
 ///
 /// A leap year is a year with 366 days. Typical years have 365 days.
 ///
 /// This doesn't used range integers and is thus `const`.
+#[inline]
 pub(crate) const fn is_leap_year_unranged(year: i16) -> bool {
-    year % 4 == 0 && (year % 100 != 0 || year % 400 == 0)
+    let d = if year % 25 != 0 { 4 } else { 16 };
+    (year % d) == 0
```
scottlamb, line 57: Why not check if `year` is an even multiple of 4/16 with masking operations instead? https://godbolt.org/z/P3xoqodfs The compiler does this sort of optimization when the dividend is constant, but in your version it's not.

Did you see a change in a benchmark here? I ran `cargo criterion` and didn't see any big changes between original / yours / my versions of `is_leap_year{,_unranged}`; not entirely sure there is a benchmark where this function is significant.

DaniPopes: The compiler does not do any sort of optimization by default :) even at -Copt-level=1 they're identical

I don't expect them to show up in any generic criterion benchmark, you can check out the [CppNow 2023 talk](https://www.youtube.com/watch?v=0s9F4QWAl-E) for more details

scottlamb: > The compiler does not do any sort of optimization by default :) even at -Copt-level=1 they're identical

I'm not sure what you're saying is identical, but it's not the two versions in my link. The assembly is substantially different.

> I don't expect them to show up in any generic criterion benchmark,

I'd suggest adding a specific one then to verify the changes are optimizations.

> you can check out the [CppNow 2023 talk](https://www.youtube.com/watch?v=0s9F4QWAl-E) at around 30 mins for more details on the optimizations done in this PR

You've pessimized the version from the talk. They did this:

```cpp
if (Y % 100 != 0)
    return Y % 4 == 0;

return Y % 16 == 0;
```

in which the division is by a constant and the compiler is easily able to strength-reduce the division. You factored the divisor into `d`, preventing it from doing so.

DaniPopes: You're comparing the generated code in debug mode, which is just wrong; you have to pass `-Copt-level=1/2/3` to Godbolt to optimize it (3 is default for --release).

Here is mine compared to the talk's: https://godbolt.org/z/T4c5xn76o

I modified it from [datealgo-rs](https://github.com/nakedible/datealgo-rs/blob/a04855a9799b0ad4c06ba6d7baa8d97760b06a57/src/lib.rs#L819-L832), which is mentioned in the author's [eaf](https://github.com/cassioneri/eaf) repository

src/util/common.rs
```diff
@@ -1,53 +1,68 @@
...
+to write `const`-compatible code, but I deemed it far too annoying to do
+so.
+
+# Algorithms
+
+Algorithms are taken from
+Neri C, Schneider L. "Euclidean affine functions and their application to calendar algorithms":
+- https://github.com/cassioneri/eaf/
+- https://www.youtube.com/watch?v=0s9F4QWAl-E
+*/
```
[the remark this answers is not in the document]
DaniPopes: Must've forgot, will do, thank

src/util/common.rs
scottlamb: > You're comparing the generated code in debug mode, which is just wrong; you have to pass -Copt-level=1/2/3 to Godbolt to optimize it (3 is default for --release).

Oh! Yeah, I thought it defaulted to `--release`-like settings, but you're right, that's a huge difference.

PR #49, opened 2024-07-26
doc: correct span format example regex

drmason13:
Just a small docs fix I noticed while browsing.

Use `s` for seconds at the end instead of `d` for days.


BurntSushi (approved):
Thanks!

BurntSushi:
FYI, the CI failure is unrelated to this PR.

PR #51, opened 2024-07-27
span: allow `ambiguous_negative_literals`

BurntSushi:
This was bad timing. The lang team just stabilized (in nightly) a new
deny-by-default lint, named `ambiguous_negative_literals`, which
triggers an error for things like `-1.hour()`. While such things can be
confusingly ambiguous in some cases, in Jiff, `-1.hour()`, `(-1).hour()`
and `-(1.hour())` are all, very intentionally, precisely equivalent.

For now we just `allow` the lint. If the lint stays, we'll likely want
to recommend allowing it in the Jiff docs.

See: https://github.com/rust-lang/rust/pull/121364

### DIFF
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -1367,6 +1367,8 @@ impl Date {
     /// # Example: negative spans are supported
     ///
     /// ```
+    /// # // See: https://github.com/rust-lang/rust/pull/121364
+    /// # #![allow(unknown_lints, ambiguous_negative_literals)]
     /// use jiff::{civil::Date, ToSpan};
     ///
     /// let d = Date::constant(2024, 3, 31);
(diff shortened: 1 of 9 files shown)


BurntSushi:
Ug. There's also a billion doc tests that will need to have this lint `allow`'d too.

src/civil/date.rs
```diff
@@ -1367,6 +1367,8 @@ impl Date {
     /// # Example: negative spans are supported
     ///
     /// ```
+    /// # // See: https://github.com/rust-lang/rust/pull/121364
+    /// # #![allow(unknown_lints, ambiguous_negative_literals)]
```
Urgau, line 1371: All of those manual allow-attr can be replaced with a single [`#![doc(test(attr(...)))]`](https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#testattr) attr: `#![doc(test(attr(allow(unknown_lints, ambiguous_negative_literals))))]`

BurntSushi: Oh wow, that's great to know! Thank you.

PR #58, opened 2024-07-29
jiff: add robust WASM support

### LINKED ISSUE #56: TimeZone::system(): support WASM as a platform
While porting Numbat to jiff (https://github.com/sharkdp/numbat/pull/511), I noticed that jiff's timezone detection doesn't work on WASM (~~currently panics with *"time not implemented on this platform"*~~). I am currently using [`iana-time-zone`](https://crates.io/crates/iana-time-zone) which supports timezone detection on wasm32 out of the box. I'm opening this ticket since [you've mentioned](https://github.com/BurntSushi/jiff?tab=readme-ov-file#platform-support) that jiff attempts to support more platforms in the future.

Here is how `iana-time-zone` does the detection for Browser/Node based environments: https://github.com/strawlab/iana-time-zone/pull/38. They basically run
```js
Intl.DateTimeFormat().resolvedOptions().timeZone
```
using `js_sys`. This is supported in [all major browsers](https://caniuse.com/mdn-javascript_builtins_intl_datetimeformat_resolvedoptions).

BurntSushi:
This PR basically makes the `wasm{32,64}-unknown-unknown` targets work
_almost_ out of the box. All you need to do is enable Jiff's new `js`
crate feature. This will cause Jiff to depend on `js-sys` and
`wasm-bindgen`. Jiff will then use Javascript APIs to determine the
current time and time zone.

This PR also includes a new long form guide, `PLATFORM.md`, which
describes Jiff's platform support in one central location. (Most
information is already in Jiff's API docs, but it's scattered in a
variety of places.)

Finally, this adds a `wasm32-unknown-unknown` test to CI courtesy of
`wasm-pack`. It just does a basic sanity check that the current time and
time zone can be retrieved.

Fixes #56

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,18 +29,6 @@ permissions:
 
 jobs:
   test:
-    env:
-      # For some builds, we use cross to test on 32-bit and big-endian
-      # systems.
-      CARGO: cargo
-      # When CARGO is set to CROSS, TARGET is set to `--target matrix.target`.
-      # Note that we only use cross on Linux, so setting a target on a
-      # different OS will just use normal cargo.
-      TARGET:
-      # Bump this as appropriate. We pin to a version to make sure CI
-      # continues to work as cross releases in the past have broken things
-      # in subtle ways.
-      CROSS_VERSION: v0.2.5
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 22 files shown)


BurntSushi:
For posterity, I had [this conversion](https://bytecodealliance.zulipchat.com/#narrow/stream/219900-wasi/topic/how.20to.20determine.20correct.20.60cfg.60.20gates.20for.20wasm.20targets.3F) with the wasmtime folks about how best to approach WASM support. Their help was instrumental in figuring out what to do here, and also in helping to educate me enough to write some half way decent WASM platform documentation.

PR #71, opened 2024-08-02
doc: minor correction: "printf"-style API

sharkdp:
I'm not a native speaker, so maybe I'm missing something. But the suggested phrasing would be easier for me to understand. I paused for a second at the "of" in *"The format string uses a "printf" style of API where conversion …"* when I read the docs. I also added a hyphen, but that can also be removed again.

### DIFF
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -2075,11 +2075,11 @@ impl Date {
     }
 }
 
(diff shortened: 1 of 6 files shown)


BurntSushi (approved):
Yeah this seems fine to me!

PR #77, opened 2024-08-03
test: disable optimizations when testing

BurntSushi:
I was getting pretty tired of the long compile times in between test
runs. I thought they seemed particularly long for debug mode. It turns
out that I had actually set `opt-level = 3` for running tests. This was
because there were a select few tests that ran quite long in debug mode.
But iteration time is important, so I trimmed down those tests and
disabled optimizations. Tests run much more quickly now after making a
change to the source code.

We also add `cargo test --profile testrelease` to CI. This is just like
normal tests, but disables `debug_assertions`. Jiff has a lot of extra
stuff going on when `debug_assertions` are enabled due to its internal
ranged integer abstraction. So it's useful to run tests without all that
extra stuff too (reflecting what is intended to be run in production).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,9 +49,6 @@ jobs:
         - build: win-msvc
           os: windows-latest
           rust: stable
-        - build: win-gnu
-          os: windows-latest
-          rust: stable-x86_64-gnu
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


BurntSushi:
The nightly CI runner has now dropped from [5m43s](https://github.com/BurntSushi/jiff/actions/runs/10231254851/job/28307003211?pr=760) to [3m7s](https://github.com/BurntSushi/jiff/actions/runs/10231483150/job/28307458102?pr=77). w00t!

PR #86, opened 2024-08-09
api: add new top-level SignedDuration type

### LINKED ISSUE #21: add more integration with `std::time::Duration`?
There are two different use cases to untangle here.

The first use case is something like this:

> I use Rust's standard library `std::time::Duration` type, and I'd like to use it directly to add and subtract elapsed/absolute time to Jiff's datetime types. I would also like to be able to _compute_ the absolute durations between datetime types using APIs like `Zoned::since`, but with `std::time::Duration` as the result returned.

The second use case is something like this:

> I like Jiff's high level `Span` data type for representing both calendar and absolute durations, but this necessarily requires a lot of overhead when performing operations. For example, adding a `Span` to a `Timestamp` is not just a simple ~single integer addition. The implementation collapses all allowed units (hours and lower) into a single integer nanosecond (an `i128`), then converts the `Timestamp` into an `i128` as a number of nanoseconds, then adds them and finally converts the integer number of nanoseconds as an `i128` back to `Timestamp`'s internal 96-bit integer representation. All of this overhead means that adding a simple count of integer seconds to a `Timestamp` is not as fast as I expect it to be.

The second use case is perhaps related to #17, but it is somewhat intertwined here, because one (but not the only) obvious way to address that is to satisfy the first use case: we provide alternative APIs that use `std::time::Duration` (hereafter just called `Duration`). This works because a `Duration` is itself just a 96-bit integer. And it always represents absolute time. That is, it is logically an integer number of nanoseconds. Therefore, adding a `Duration` to something like a `Timestamp` should be a very fast operation. Certainly faster than the existing `Span` focused APIs that we have today.

At present, I believe it is correct to say that `Duration` appears in the public API exactly zero times. This was not an accident, because I was unsure of how to integrate it. I think there are two challenges:

The first is that `Duration` is unsigned. There is some discussion in the [issue tracking the addition of `Duration` (and `SystemTime`) to the standard library](https://github.com/rust-lang/rust/issues/29866), and the [related RFC](https://rust-lang.github.io/rfcs/1288-time-improvements.html) has this to say:

> Finally, many APIs that take a Duration can only do something useful with positive values. For example, a timeout API would not know how to wait a negative amount of time before timing out. Even discounting the possibility of coding mistakes, the problem of system clock time travel means that programmers often produce negative durations that they did not expect, and APIs that liberally accept negative durations only propagate the error further.

I was on libs (now libs-api) at the time, and I'm not sure if this was a mistake or not. Maybe the world really does need both unsigned and signed durations. The main issue with an unsigned duration type in Jiff is that it isn't nearly as flexible. For example, if you do `ts1.until(ts2)` and `ts1` refers to a point in time after `ts2`, then Jiff will happily return a negative span. But what happens if you want to return a `Duration`? What is the result? One choice is to just return the duration between them such that `ts1.until(ts2)` and `ts2.until(ts1)` are equivalent. The other is to return an error. I would probably lean towards "return an error"... This doesn't come up as much with things like `ts.checked_add(duration)` since there is a `ts.checked_sub(duration)`, and because it's up to caller in this case to apply the right operation. It isn't on Jiff to interpret the duration as anything other than unsigned.

The second problem with `Duration` integration is that it's not clear how to, well, actually integrate it. Does this mean Jiff should grow `checked_add_duration` and `since_duration` methods? I think so... And I think this is probably a requirement to satisfying the second use case above, even if we use a different duration type. I do personally find the names with a `_duration` suffix quite clunky, but perhaps it's fine because these will be considered "niche" APIs IMO.

It would perhaps be possible to satisfy the first use case without satisfying the second while simultaneously _not_ adding a bunch of clunky `_duration` APIs to all of the datetime types. This could be accomplished by providing APIs to convert between a `Span` and a `Duration`. So callers would still need to use `Span`, but they would at least have a paved path for using `Duration`.

I think the main alternative point in the design space here is for Jiff to define its _own_ absolute duration type that is signed. We'd still need things like `checked_add_duration` and `since_duration`, but the awkwardness with an unsigned duration type would go away. And then once we have our own absolute duration type, we could define conversions to and from std's unsigned `Duration` type. And we could deal with the peculiarities of signedness in a very focused and deliberate way.

BurntSushi:
This PR adds a new top-level `SignedDuration` type to Jiff. This PR also
does the work to integrate it into all Jiff datetime APIs. For example,
`Zoned::checked_add` now accepts a `jiff::Span`, `jiff::SignedDuration`
or a `std::time::Duration`. To achieve this, we add a new target type,
`ZonedArithmetic`, with `From` trait implementations for each of the
corresponding duration types. We then changed the `span: Span`
parameter type to `duration: impl Into<ZonedArithmetic>`. We also add
new `Arithmetic` target types for each datetime type, for example,
`TimestampArithmetic`.

This design was chosen over using one target type shared across all
datetime types to allow for future extensibility. For example, maybe in
the future we want to add extra options to `Zoned::checked_add`. We can
now do that via `ZonedArithmetic` in a way that is distinct from other
datetime types. Moreover, this also avoids defining a new type that
*looks* a lot like a duration. Instead, the target types are purpose
built and constrained to arithmetic operations involving datetimes and
durations.

The main motivation for this was to facilitate faster arithmetic
operations in cases where folks need them and tighter integration with
the standard library `std::time::Duration` type. The `Span` type is
still the primary duration type in Jiff that almost everyone should be
using most of the time.

Closes #21

### DIFF
--- a/src/signed_duration.rs
+++ b/src/signed_duration.rs
@@ -117,8 +117,8 @@ use crate::util::libm::Float;
 /// to `i64::MAX`, which is less than `u64::MAX`.
 /// * There are some additional APIs that don't make sense on an unsigned
 /// duration, like [`SignedDuration::abs`] and [`SignedDuration::checked_neg`].
-/// * A [`SignedDuration::until`] routine is provided as a replacement for
-/// [`std::time::SystemTime::duration_since`], but with signed durations.
+/// * A [`SignedDuration::system_until`] routine is provided as a replacement
+/// for [`std::time::SystemTime::duration_since`], but with signed durations.
 /// * Constructors and getters for units of hours and minutes are provided,
 /// where as these routines are unstable in the standard library.
 /// * Unlike the standard library, this type implements the `std::fmt::Display`
--- a/src/signed_duration.rs
+++ b/src/signed_duration.rs
@@ -170,6 +170,113 @@ use crate::util::libm::Float;
 /// (as long as they're within Jiff's limits), including durations with units
 /// of years, months, weeks and days. A `SignedDuration`, by contrast, only
 /// supports units up to and including hours.
+///
+/// # Integration with datetime types
+///
+/// All datetime types that support arithmetic using [`Span`](crate::Span) also
+/// support arithmetic using `SignedDuration` (and [`std::time::Duration`]).
+/// For example, here's how to add an absolute duration to a [`Timestamp`]:
+///
+/// ```
+/// use jiff::{SignedDuration, Timestamp};
+///
+/// let ts1 = Timestamp::from_second(1_123_456_789)?;
+/// assert_eq!(ts1.to_string(), "2005-08-07T23:19:49Z");
+///
+/// let duration = SignedDuration::new(59, 999_999_999);
+/// // Timestamp::checked_add is polymorphic! It can accept a
+/// // span or a duration.
+/// let ts2 = ts1.checked_add(duration)?;
+/// assert_eq!(ts2.to_string(), "2005-08-07T23:20:48.999999999Z");
+///
+/// # Ok::<(), Box<dyn std::error::Error>>(())
+/// ```
+///
+/// The same API pattern works with [`Zoned`], [`DateTime`], [`Date`] and
+/// [`Time`].
+///
+/// # Interaction with daylight saving time and time zone transitions
+///
+/// A `SignedDuration` always corresponds to a specific number of nanoseconds.
+/// Since a [`Zoned`] is always a precise instant in time, adding a `SignedDuration`
+/// to a `Zoned` always behaves by adding the nanoseconds from the duration to
+/// the timestamp inside of `Zoned`. Consider `2024-03-10` in `US/Eastern`.
+/// At `02:00:00`, daylight saving time came into effect, switching the UTC
+/// offset for the region from `-05` to `-04`. This has the effect of skipping
+/// an hour on the clocks:
+///
+/// ```
+/// use jiff::{civil::date, SignedDuration};
+///
+/// let zdt = date(2024, 3, 10).at(1, 59, 0, 0).intz("US/Eastern")?;
+/// assert_eq!(
+///     zdt.checked_add(SignedDuration::from_hours(1))?,
+///     // Time on the clock skipped an hour, but in this time
+///     // zone, 03:59 is actually precisely 1 hour later than
+///     // 01:59.
+///     date(2024, 3, 10).at(3, 59, 0, 0).intz("US/Eastern")?,
+/// );
+/// // The same would apply if you used a `Span`:
+/// assert_eq!(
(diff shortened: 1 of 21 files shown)


src/signed_duration.rs
```diff
@@ -0,0 +1,2189 @@
...
+        SignedDuration::new_unchecked(secs, nanos)
+    }
+
+    /// Creates a new `SignedDuration` from the given number of hours. Every
+    /// hour is exactly `3,600` seconds.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of hours, after being converted to nanoseconds,
+    /// overflows the minimum or maximum `SignedDuration` values.
```
tisonkun, line 568: Shall we then provide a fallible constructor like `try_from_hours`? So does `from_mins`.

BurntSushi: There aren't fallible constructors for any units. That matches std. The design philosophy is for SignedDuration to match std as closely as possible. I'm open to adding fallible constructors, but I want to see compelling use cases.

tisonkun: Thanks for your explanation. I noticed that the upstream APIs are unstable also, so let's see what's the consensus - https://github.com/rust-lang/rust/issues/120301/

Both fallible constructor and ambiguous semantic are thrown. I noticed an argument that `hours` and `minutes` are not supported for leap seconds and more ([ref](https://rust-lang.github.io/rfcs/1040-duration-reform.html)), but in jiff we typically consider days and above are ambiguous? Is there certain different consideration?

BurntSushi: Sorry, I don't understand your question. Jiff specifically and intentionally doesn't support leap seconds. And whether leap seconds, even if supported, should impact the definition of hours or minutes is unclear.

tisonkun: > doesn't support leap seconds

OK. Then it looks some concerns from the [Rust RFC 1040](https://rust-lang.github.io/rfcs/1040-duration-reform.html) proposer.

> even if supported, should impact the definition of hours or minutes is unclear

Yeah .. So the reason jiff don't support days and above to `SignedDuration`, is that:

* How many days in a year is indeterminate (leap year)
* How many days in a month is indeterminate (of course)
* How many hours in a day is indeterminate (I'm not sure but perhaps daylight savings things?)
* How many hours in a week is indeterminate (because of day -> hours is indeterminate)

Is this understanding properly?

BurntSushi: Yes

BurntSushi: I would definitely encourage you to read more of Jiff's docs. And especially the examples. They'll demonstrate a lot of these weird edge cases.

tisonkun: > I would definitely encourage you to read more of Jiff's docs. And especially the examples. They'll demonstrate a lot of these weird edge cases.

Sure .. Thanks for providing such detailed docs.

PR #97, opened 2024-08-14
use threads to update default time zone cache asynchronously

### LINKED ISSUE #96: Optimize the performance of obtaining the default time zone
# Background

`TimeZone::system()` obtains the default time zone, and it may be called frequently by users. 

Especially, `Zoned::now` will call it. In some systems, `Zoned::now` is frequently used to obtain the current time. For example, log system will call it for each log item.

# How long does `TimeZone::system()` need? and why?

After run the benchmark on my Mac, I found `TimeZone::system()` needs `61.665 ns`, its cost mainly consists of 3 parts:

1. TimeZone.clone[1] costs ***9.3 ns***, it clones the TimeZone from cache.
2. Request the read lock[2] costs ***16.88 ns***. (The read lock of cache.)
3. Obtain current time via `Instant::now`[3] costs ***37.3 ns***, the Instant is used for checking whether the cache is expired.

So the `Instant::now`(part 3) takes most of time.

# The idea of optimization

Don't call `Instant::now` during call `TimeZone::system()`, and we could start a light async thread to update the default time zone cache periodically.

Note: if the default time zone is obtained by async thread, we could update it more frequently. For example: update the default TTL from 5 minutes to 20 seconds or less.

[1] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L119
[2] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L104
[3] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L106

1996fanrui:
close #96 

# Background

`TimeZone::system()` obtains the default time zone, and it may be called frequently by users. 

Especially, `Zoned::now` will call it. In some systems, `Zoned::now` is frequently used to obtain the current time. For example, log system will call it for each log item.

# How long does `TimeZone::system()` need? and why?a

After run the benchmark on my Mac, I found `TimeZone::system()` needs `61.665 ns`, its cost mainly consists of 3 parts:

1. TimeZone.clone[1] costs ***9.3 ns***, it clones the TimeZone from cache.
2. Request the read lock[2] costs ***16.88 ns***. (The read lock of cache.)
3. Obtain current time via `Instant::now`[3] costs ***37.3 ns***, the Instant is used for checking whether the cache is expired.

So the `Instant::now`(part 3) takes most of time.

# The idea of optimization

Don't call `Instant::now` during call `TimeZone::system()`, and we could start a light async thread to update the default time zone cache periodically.

Note: if the default time zone is obtained by async thread, we could update it more frequently. For example: update the default TTL from 5 minutes to 20 seconds or less.

[1] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L119
[2] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L104
[3] https://github.com/BurntSushi/jiff/blob/c659069a2765a5e28ba503ea37e8a2eee4b10a92/src/tz/system/mod.rs#L106

This PR is a draft PR due to it includes some temporary benchmark code or use case.

I will polish this PR if the solution makes sense.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ keywords = ["date", "time", "temporal", "zone", "iana"]
 edition = "2021"
 exclude = ["/.github", "/tmp"]
 autotests = false
-autoexamples = false
 rust-version = "1.70"
 
 [workspace]
@@ -23,7 +22,6 @@ members = [
   "jiff-cli",
   "jiff-tzdb",
   "jiff-tzdb-platform",
-  "examples/*",
 ]
 
 # Features are documented in the "Crate features" section of the crate docs:
--- a/bench/Cargo.toml
+++ b/bench/Cargo.toml
@@ -19,6 +19,11 @@ name = "jiff-bench"
 harness = false
 path = "src/bench.rs"
 
+[[bench]]
+name = "default_time_zone_benchmark"
+harness = false
+path = "src/default_time_zone_benchmark.rs"
+
 [dependencies]
 criterion = "0.5.1"
 jiff = { version = "0.1.0", path = ".." }
--- a/bench/src/default_time_zone_benchmark.rs
+++ b/bench/src/default_time_zone_benchmark.rs
@@ -1,0 +1,35 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use jiff::tz::TimeZone;
+use std::sync::RwLock;
+use std::time::Instant;
+
+fn default_time_zone_benchmark(c: &mut Criterion) {
+    c.bench_function("Get default TimeZone::system()", |b| {
+        b.iter(|| {
+            TimeZone::system();
+        })
+    });
+
+    c.bench_function("Clone a TimeZone", |b| {
+        let tz = TimeZone::system();
+        b.iter(|| {
+            let _ = tz.clone();
+        })
+    });
+
+    c.bench_function("Read lock", |b| {
+        let cache: RwLock<usize> = RwLock::new(0);
+        b.iter(|| {
+            let a = cache.read().unwrap();
+        })
+    });
+
+    c.bench_function("Instant::now", |b| {
+        b.iter(|| {
+            let _ = Instant::now();
+        })
+    });
+}
+
+criterion_group!(benches, default_time_zone_benchmark);
+criterion_main!(benches);
--- a/examples/default_time_zone_demo/main.rs
+++ b/examples/default_time_zone_demo/main.rs
@@ -1,0 +1,36 @@
+use jiff::tz::TimeZone;
+use std::thread::sleep;
+use std::time::Duration;
+
+fn main() {
+    // The time zone is updated by async thread each 20 seconds.
+    // And we get the default time zone continuously,
+    // so we can see the log: `cache is still using, so update the tz.`
+    for i in 1..=50 {
+        sleep(Duration::from_secs(1));
+        TimeZone::system();
+        println!("round 1------{i}");
+    }
+
+    // Stop the get the default time zone for a while, so we can see the log :
+    // `cache is not used so far, so stop this thread.`
+    for i in 1..=35 {
+        sleep(Duration::from_secs(1));
+        println!("round 2------{i}");
+    }
+
+    // Get the default time zone again, we can see
+    // `try_update_time_zone` log to start the thread again.
+    // And see the `cache is still using, so update the tz.` log later.
+    for i in 1..=50 {
+        sleep(Duration::from_secs(1));
+        TimeZone::system();
+        println!("round 3------{i}");
+    }
+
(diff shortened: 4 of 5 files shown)


BurntSushi:
I don't think this is the way to go. We shouldn't be starting threads inside a low level library like this. Or at least, I would want a strong ecosystem precedent for it.

1996fanrui:
> I don't think this is the way to go. We shouldn't be starting threads inside a low level library like this. Or at least, I would want a strong ecosystem precedent for it.

I also think it's not suitable, so only submit a draft PR to check with you, thanks for your quick feedbacck. 

Do you think `Zoned::now`  is frequently used to obtain the current time for the users of jiff? 

An alternative solution is using SystemTIme to check the TTL instead of Instant. Because  `Zoned::now` will call `SystemTime::now()` and `TimeZone::system()`.  If we use SystemTIme to check the TTL, and pass `SystemTime::now()` to `TimeZone::system()`, `TimeZone::system()` doesn't need to call `Instant::now`. It could save some time, WDYT?

Of course, if user calls `TimeZone::system()` directly, we still need to call `SystemTime::now()`  to check TTL.

Also, I see the TTL is 5 miuntes by default, so the time error of `SystemTime::now()` should be within the tolerance range.

BurntSushi:
> Also, I see the TTL is 5 miuntes by default, so the time error of `SystemTime::now()` should be within the tolerance range.

I think this is kinda missing the point of using monotonic time instead of system time on its own. It's not really about "time error." The system time can change _arbitrarily_ and at _any time_. It's not that system time has some kind of error built into it necessarily. It's just that it's configurable. So if the code _relies_ on time always increasing (or rather, being non-decreasing), then monotonic time is very useful to that end.

With that said, I do actually think this is an interesting idea. I don't think we can just get rid of the monotonic time check completely, but we might be able to manage something here because we are dealing with caching. In particular, it's _always_ okay to invalidate the cache. The goal is to just do it infrequently. But if there's an "odd" case where we can't know if the cache is fresh or not, we can fall back to monotonic time.

1. Whenever the cache is invalidated, we should record the expiration as _both_ a monotonic time and a system time.
2. If no current `SystemTime` is available when requesting the system time zone, use the monotonic time expiration.
3. Otherwise, compare the current `SystemTime` with the expiration's `SystemTime`:
  1. If the difference is greater than the TTL (in either direction), then use monotonic time.
  2. If the current `SystemTime` is less than the expiration `SystemTime`, treat the cache as fresh.
  3. Otherwise, use monotonic time.

I think this is not fully bullet proof. In particular, if the `SystemTime` is continually reset to a point before the expiration `SystemTime` that is within the TTL, then the cache will always be fresh. To work around this, I think we can keep a count of the number of times we make a caching decision, consecutively, _without_ using monotonic time. If this count gets too large, then we automatically fallback to monotonic time. This will prevent us from getting stuck in cases of pathological system time behavior.

I think that should actually do it? I don't immediately see a way for this to go wrong. (Famous last words.)

1996fanrui:
Thanks @BurntSushi for the detailed and valuable feedback!

>   1. Whenever the cache is invalidated, we should record the expiration as _both_ a monotonic time and a system time.
>   2. If no current `SystemTime` is available when requesting the system time zone, use the monotonic time expiration.
>   3. Otherwise, compare the current `SystemTime` with the expiration's `SystemTime`:
>   4. If the difference is greater than the TTL (in either direction), then use monotonic time.
>   5. If the current `SystemTime` is less than the expiration `SystemTime`, treat the cache as fresh.
>   6. Otherwise, use monotonic time.

Overall LGTM.

> I think we can keep a count of the number of times we make a caching decision, consecutively, _without_ using monotonic time. If this count gets too large, then we automatically fallback to monotonic time. 

Recording a count of using `SystemTime` and fallback to monotonic time sounds make sense to me. 

I have a question: if we find that `SystemTime` is correct after falling back to monotonic time once, should we continue to use `SystemTime`?

BurntSushi:
> I have a question: if we find that SystemTime is correct after falling back to monotonic time once, should we continue to use SystemTime?

I'm not sure. So pick whichever is simplest to implement. Probably just continue using `SystemTime`.

BurntSushi:
I'm going to close this out because I think this particular avenue for optimization is not the right path to go down. But if you want to try the other idea I put forward, it's probably best to do that in a separate PR. I may also try it myself at some point too. Thanks for the effort!

PR #106, opened 2024-08-18
cargo: permit windows-sys >=0.52.0 and <=0.59.*

BurntSushi:
This matches a similar change made by @ChrisDenton for winapi-util in:
https://github.com/BurntSushi/winapi-util/pull/19

The benefit is that this should hopefully put less constraints on the
ecosystem and to enable fewer copies of windows-sys to appear in
dependency trees.

I am somewhat miffed that this is necessary. IMO, the windows-sys
maintainers should find a way to make a stable release with infrequent
or ~zero breaking change releases.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,7 +87,7 @@ serde = { version = "1.0.203", optional = true }
 jiff-tzdb-platform = { version = "0.1.0", path = "jiff-tzdb-platform", optional = true }
 
 [target.'cfg(windows)'.dependencies.windows-sys]
-version = "0.52.0"
+version = ">=0.52.0, <=0.59.*"
 default-features = false
 features = ["Win32_Foundation", "Win32_System_Time"]
 optional = true


ChrisDenton:
> IMO, the windows-sys maintainers should find a way to make a stable release with infrequent or ~zero breaking change releases.

The windows-sys crate is auto-generated from upstream [metadata](https://github.com/microsoft/win32metadata) so only has so much control. And people do want fixes for issues so zero breaking changes won't be realistic until the metadata stabilizes and is fully reliable. I don't think infrequent release would solve the issue either. Whether it's a breaking release every six months or two years, there's going to be a point where the ecosystem is split for a long time due to the very different rates at which crates update their dependencies (almost immediately, via a bot, or a year or two later when someone happens to notice the outdated dependency).

There is however another option. Bespoke bindings can be generated using the [windows-bindgen](https://crates.io/crates/windows-bindgen) crate. The only public dependency is the `windows-targets` crate which should rarely (if ever) require an update. You can see a quick example of using it in the crate's readme. The windows-rs repository uses it for some crates. E.g. see [strings.txt].

BurntSushi:
Thank you! I'll probably move to `windows-targets` at some point, since there are very small interaction points (at least currently) with the Windows API.

PR #114, opened 2024-08-23
api: add options for controlling precision

### LINKED ISSUE #92: Allow configure nanosecond's precision
Background - https://github.com/cratesland/logforth/pull/40#discussion_r1712993395

Currently, the default formatter can cause unaligned nanosecond:

```
2024-08-11T13:47:37.947497+00:00[UTC] ERROR simple_stdio: examples/simple_stdio.rs:32 Hello error!
2024-08-11T13:47:37.98353+00:00[UTC]  WARN simple_stdio: examples/simple_stdio.rs:33 Hello warn!
2024-08-11T13:47:37.983557+00:00[UTC]  INFO simple_stdio: examples/simple_stdio.rs:34 Hello info!
2024-08-11T13:47:37.983579+00:00[UTC] DEBUG simple_stdio: examples/simple_stdio.rs:35 Hello debug!
2024-08-11T13:47:37.983602+00:00[UTC] TRACE simple_stdio: examples/simple_stdio.rs:36 Hello trace!
```

@BurntSushi suggests we can make it configurable at both:

* [`jiff::fmt::temporal::DateTimePrinter`](https://docs.rs/jiff/latest/jiff/fmt/temporal/struct.DateTimePrinter.html)
* https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.precision

Record for any improvement chance.

BurntSushi:
This PR introduces new options for controlling the precision
of fractional seconds when printing `Zoned`, `Timestamp`,
`civil::DateTime` or `civil::Time` values. This is principally exposed
via `jiff::fmt::temporal::DateTimePrinter::precision`, but it's also
available via the standard library's formatting machinery. For example,
if `zdt` is a `jiff::Zoned`, then `format!("{zdt:.6}")` will format
it into a string with microsecond precision, even if its fractional
component is zero.

This is useful when one wants a datetime to use a "fixed width" format.
Or at least, as close to one as possible. For `Zoned` in particular, a
fixed width format is somewhat difficult to accomplish because of the
variable length IANA time zone identifier. But if the time zone
identifier is the same for all `Zoned` values in a particular context,
then setting the precision will provide fixed width. (Unless the years
are negative.)

Closes #92

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+0.1.9 (2024-08-23)
+==================
+This release introduces new options for controlling the precision
+of fractional seconds when printing `Zoned`, `Timestamp`,
+`civil::DateTime` or `civil::Time` values. This is principally exposed
+via `jiff::fmt::temporal::DateTimePrinter::precision`, but it's also
+available via the standard library's formatting machinery. For example,
+if `zdt` is a `jiff::Zoned`, then `format!("{zdt:.6}")` will format
+it into a string with microsecond precision, even if its fractional
+component is zero.
+
+Enhancements:
+
+* [#92](https://github.com/BurntSushi/jiff/issues/92):
+Support setting the precision of fractional seconds when printing datetimes.
+
+
 0.1.8 (2024-08-19)
 ==================
 This releases fixes a build error in Jiff's `alloc`-only configuration. This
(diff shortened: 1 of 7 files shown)


BurntSushi:
This PR is on crates.io in `jiff 0.1.9`.

PR #117, opened 2024-08-27
util/libm: replace `force_eval!` by `std::hint::black_box`

martinvonz:
During an internal review of unsafe code at work, @cramertj noticed that the the `force_eval!` can be replaced by `std::hint::black_box`, removing the need for unsafe Rust.

### DIFF
--- a/src/util/libm.rs
+++ b/src/util/libm.rs
@@ -24,12 +24,6 @@ pub(crate) trait Float {
     fn fract(self) -> Self;
 }
 
-macro_rules! force_eval {
-    ($e:expr) => {
-        core::ptr::read_volatile(&$e)
-    };
-}
-
 const TOINT64: f64 = 1. / f64::EPSILON;
 
 impl Float for f64 {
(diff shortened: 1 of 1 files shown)


cramertj:
Related: https://github.com/rust-lang/libm/pull/288

src/util/libm.rs
```diff
@@ -129,7 +119,7 @@ impl Float for f64 {
             return x;
         }
         unsafe {
```
BurntSushi, line 121: I think you forgot to remove this `unsafe`?

BurntSushi (changes requested):
It looks like one `unsafe` was missed.

Otherwise, insomuch as I can judge this, I think this LGTM. However, @Amanieu did say this:

> To work around FP precision issues on 32-bit x86 which uses 80-bit float precision internally. This is unfortunately an unavoidable issue on x86 since it has pretty broken FP support.

Which seems to me like it's relying on volatile reads for correctness. But my understanding is that relying on `black_box` for correctness is bad juju. I don't understand enough about the problem domain here to adjudicate though. I'd be inclined to take a reactive approach here and merge, and revert if folks notice problems in practice.

src/util/libm.rs
martinvonz: Yep. Fixed.

BurntSushi (approved):

Issue #7, opened 2024-04-06
about leap seconds


BurntSushi:
### What should you do if you want leap second support in Jiff?

The most effective thing you can do is leave a comment in this issue with a _detailed_ account of why you want leap second support. Leaving a comment like, "I need to-the-second precision when calculating differences between instants" is not sufficient. You should say _why_ you need that level of precision, and especially, what would happen if you didn't have it. You should also share what your current solution is and any relevant "industry standards" in your domain. (So for example, if your use case is, "I need to do astronomical math," then it would be nice to say why you can't or don't want to use a specialized library for this purpose.) The most important thing you can do is **connect specific support of leap seconds to an end user experience.** Nebulous "I want it because it's correct" comments are not especially helpful.

The rest of this issue does a deep dive on what leap second support means and various API designs.

-----

**NOTE:** When I originally wrote this, `Timestamp` was called `Instant`. I tried replacing references to `Instant` with `Timestamp` below, but a timestamp is still referred to as an "instant" in places. This concept is still correct. A `Timestamp` _is_ an instant. The only thing that changed was the type name in Jiff.

-----

[This required background reading](http://www.madore.org/~david/computers/unix-leap-seconds.html) gives a nice overview of the relationship between different time scales. (This also makes a proposal at the end, but that's not particularly relevant here. The content leading up to the proposal though I found quite helpful in understanding things.)

Here is a bit [more background reading](https://unix.stackexchange.com/questions/283164/unix-seconds-tai-si-seconds-leap-seconds-and-real-world-code), which serves as a nice way of observing that pretty much everybody gets this wrong in some way. And the reason why we get it wrong pretty much comes down to the fact that we like to represent time as "seconds since an epoch."

And finally, this [reference about Time Standards](https://geometrian.com/programming/reference/timestds/index.php) I found to be a bit ranty, but also very informative. It's nice because it starts from first principles and builds up everything from there.

# Initial thoughts and ecosystem comparison

When I initially started developing Jiff, I had a goal to provide support for leap seconds. The failure to support leap seconds is a common indictment of datetime libraries, and I wanted to avoid falling into that trap. Roughly speaking, "support" in my mind looked like this. And note that when I say `Timestamp`, it also applies to `Zoned` (but not necessarily `civil::DateTime`).

[...]

PR #134, opened 2024-09-19
Tweak docs to update Timestamp definition to current behaviour

Plecra:
~~The end goal with this PR is to adjust the documentation to make the semantics of the library as clear as possible.~~

~~I've started with just a tiny change to set a foundation, updating the documentation comment to define Timestamp wrt UNIX time.~~

citations, since my only msgs so far are on discord (I believe Timestamp = Unix is currently intentional, this is just a mistake in docs):

- `now` uses the value constructed by `duration_since(EPOCH)`, which is [defined as POSIX time](https://doc.rust-lang.org/stable/std/time/struct.SystemTime.html#associatedconstant.UNIX_EPOCH)
  > duration_since(UNIX_EPOCH).unwrap().as_secs() returns the number of non-leap seconds since the start of 1970 UTC. This is a POSIX time_t (as a u64), and is the same time representation as used in many Internet protocols.
  this can also be tested experimentally by printing the `Timestamp::now().as_nanos()`
- `Display for Timestamp` prints the corresponding date to an input unix time.
  this can also be tested by printing `format!("{}", Timestamp::new(24 * 60 * 60 * (365 * 30 + (30 / 4)) , 0))` to `2000-01-01T00:00:00Z`, which is 22 seconds faster than it should be

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -23,7 +23,7 @@ use crate::{
 /// An instant in time represented as the number of nanoseconds since the Unix
 /// epoch.
 ///
-/// A timestamp is always in the UNIX timescale.
+/// A timestamp is always in the UNIX timescale with a UTC offset of zero.
 ///
 /// To obtain civil or "local" datetime units like year, month, day or hour, a
 /// timestamp needs to be combined with a [`TimeZone`] to create a [`Zoned`].
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -23,7 +23,7 @@ use crate::{
 /// An instant in time represented as the number of nanoseconds since the Unix
 /// epoch.
 ///
-/// A timestamp is always in UTC.
+/// A timestamp is always in the UNIX timescale.
 ///
 /// To obtain civil or "local" datetime units like year, month, day or hour, a
 /// timestamp needs to be combined with a [`TimeZone`] to create a [`Zoned`].
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -23,7 +23,7 @@ use crate::{
 /// An instant in time represented as the number of nanoseconds since the Unix
 /// epoch.
 ///
-/// A timestamp is always in the UNIX timescale with a UTC offset of zero.
+/// A timestamp is always in the Unix timescale with a UTC offset of zero.
 ///
 /// To obtain civil or "local" datetime units like year, month, day or hour, a
 /// timestamp needs to be combined with a [`TimeZone`] to create a [`Zoned`].


Plecra:
On https://github.com/BurntSushi/jiff/issues/7#issuecomment-2359423476, I wanna make the most compelling case I can 😁 

Which part of this should I elaborate on?

- "Jiff is a datetime library for Rust that encourages you to jump into the pit of success", so I'm assuming the API should be structured in a way that makes it easy to find the right API when you want to do something
- It is common for users to write the `let start = now(); f(); now() - start` pattern
- The current API presents `Timestamp` as entirely appropriate (`now()` and `Sub`) for the measurement pattern
- But this implementation is obviously wrong
  - in that it will return a completely incorrect result in normal operating conditions 
  - *and* 1 second is a massive error for plenty of cases of this measurement pattern


Youve said that

> I also think that a lot of the trouble and bugs that come from leap seconds aren't necessarily because of datetime libraries not handling them, but rather, because programmers have assumed that their system time is monotonically increasing. Jiff doesn't provide any monotonic guarantees because it's a library for dealing with human time.

so this is entirely fine semantically, but I do not agree that it's a "pit of success", the library immediately lures naive users down the wrong path by presenting them with the classic failure mode. We *cant* write code that calculates deltas because we can't "assume that our system time is monotonically increasing", but jiff offers that incorrect behaviour via the most accessible API!

BurntSushi:
> so this is entirely fine semantically, but I do not agree that it's a "pit of success", the library immediately lures naive users down the wrong path by presenting them with the classic failure mode. We cant write code that calculates deltas because we can't "assume that our system time is monotonically increasing", but jiff offers that incorrect behaviour via the most accessible API!

Pit of success is a pithy goal of the library. It's a design principle. Jiff does not and will not meet that goal in every case, and it will sometimes be in tension with other design goals. You seem to be suggesting that Jiff not offer a way to ask for the current time and provide an easy way to calculate the difference between two times. Like, this seems completely bonkers to me.

src/timestamp.rs
```diff
@@ -23,7 +23,7 @@ use crate::{
 /// An instant in time represented as the number of nanoseconds since the Unix
 /// epoch.
 ///
-/// A timestamp is always in the UNIX timescale.
+/// A timestamp is always in the UNIX timescale with a UTC offset of zero.
```
BurntSushi, line 26:
```suggestion
/// A timestamp is always in the Unix timescale with a UTC offset of zero.
```

BurntSushi (changes requested):

BurntSushi (approved):

Issue #50, opened 2024-07-26
add ecosystem integration to jiff? (e.g., optional dependencies on `icu`, `diesel`, `sqlx`, etc)


BurntSushi:
In #4 and in other places, I've resisted adding "ecosystem integration" into Jiff. Not necessarily because I expected others to instead depend on Jiff necessarily right away, but that _eventually_, as folks started using Jiff (hopefully), it would make sense for them to add a Jiff option.

But I think I should probably flip this script. Getting Jiff adopted is going to be difficult, and I don't know if it has any broadly applicable "killer feature" that will convince folks to jump over from other datetime libraries. (I think Jiff's `Span` is really amazing, but I haven't figured out the best way of communicating that to a broad audience yet.) I think this means that the best path to Jiff adoption is to reduce friction as much as possible.

I get the sense that folks rely heavily on other crates providing integration points for their chosen datetime types of choice. In particular, it is often the case that converting from one type to another is actually quite tricky. Even tricky enough that [I myself get tripped up](https://github.com/sfackler/rust-postgres/pull/1164). So expecting users to just hand-roll these sorts of integrations is probably pretty lofty.

So... I think one way to reduce friction is for Jiff to provide these integrations. One of my initial reservations was introducing public dependencies on crates that are "less stable/mature." While Jiff is currently brand new and I expect to do some breaking change releases, I do hope that in about ~1 year's time, we'll have a Jiff 1.0 and that I'll commit to that for the foreseeable future with no breaking changes. However, I actually don't think this is a huge deal. We can just include the major/minor version of the crate in the feature name itself, [just like `rust-postgres` does](https://github.com/sfackler/rust-postgres/blob/d2634a46b5bc1de5176f0766626e7a9f6757f362/postgres-types/Cargo.toml#L53-L54).

I have two concerns with this approach. That is, if I start this way by introducing dependencies on crates like `diesel`, then:

1. Does this mean it will become very difficult, if not impossible, to coordinate a switch where instead of Jiff providing the integration points, the higher level libraries do it instead?
2. A datetime library ought to appear "low" in the dependency tree. It is intended to be a building block for other systems. So if Jiff starts depending on "bigger" systems like `diesel` (`diesel` is just an example, this applies to most crates I think), then what happens if those bigger systems actually need to depend on a datetime library? Recursive dependencies are a no-go.

[...]

PR #141, opened 2024-10-11
feat: bootstrap jiff-sqlx development 

tisonkun:
This refers to #50.

Firstly, I considered to add a separated `sqlx-jiff` method, but soon I found that due to Rust's orphan rules, implementing sqlx's trait for `Timestamp` and other types requires the code to be located at the same crate (`jiff`). Otherwise we need an extra wrapper struct like:

```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
pub struct Timestamp(pub jiff::Timestamp);
```

I prefer to implement code in `jiff` and gate it with feature flags.

@BurntSushi This demonstrates the integration implementation direction. I'll complete other parts when we agree on this direction.


BurntSushi:
https://github.com/BurntSushi/jiff/issues/50#issuecomment-2253457694 discusses this. And the orphan rule is taken into account there. The way to do this integration, short of getting it merged upstream in `sqlx` itself, is with a wrapper type in a `jiff-sqlx` crate. The comment linked explains how to do this.

Jiff is a "low level" crate. Having it depend on "high level" crates like `sqlx` would lead to all sorts of problems, especially if and when the time ever comes for `sqlx` to add Jiff integration on their end. It's really a non-starter.

sqlx-jiff/src/lib.rs
```diff
@@ -0,0 +1,34 @@
+use serde::{Deserialize, Serialize};
+use std::ops::{Deref, DerefMut};
+
+#[cfg(feature = "postgres")]
+mod postgres;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Timestamp(pub jiff::Timestamp);
```
tisonkun, line 8: I'm not sure if `(pub jiff::Timestamp)` is better or if I should add a `to_jiff` method + a trait with a `to_sqlx` method, as described in https://github.com/BurntSushi/jiff/issues/50#issuecomment-2253457694.

Seems that the wrapper will always wrap exactly `jiff::Timestamp` and if we `pub` it, both construct and retrieve are convenient.

tisonkun (commented):
@BurntSushi Yeah .. Adopt the other way now.

sqlx-jiff/src/lib.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+impl From<jiff::Timestamp> for Timestamp {
+    fn from(ts: jiff::Timestamp) -> Self {
+        Self(ts)
+    }
+}
+
+impl Deref for Timestamp {
```
tisonkun, line 22: `impl AsRef` also?

sqlx-jiff/src/lib.rs
BurntSushi: Definitely not `pub` it. And even if you did, we'd still want the extension trait and the `to_jiff` method.

I'd encourage you to check out my libraries. I think there are approximately zero instances of a `pub` field like this. There are use cases for them in library crates (`std` has a few of them), but they are pretty rare.

sqlx-jiff/src/lib.rs
BurntSushi: I'm not sure how far we want to go here. In theory, there are a lot of different traits we might want to implement. In practice, I'm not sure how useful they all are or whether we even want to encourage their use. I'd start with a minimal useful set for now.

I'm actually not even sure about using `Deref` here. I'd leave that out for now.

BurntSushi:
Also, the crate should be called `jiff-sqlx`, since it lives with `jiff`.

sqlx-jiff/src/lib.rs
tisonkun: OK. Updated.

sqlx-jiff/src/lib.rs
tisonkun: OK. Then I'd revoke `#[derive(PartialEq, Eq, Serialize, Deserialize)]` also but just make it an opaque wrapper type. Users should always call `to_jiff` and use `jiff::Timestamp`'s method.

tisonkun:
Other type mapping to be wrapped once we agree on the pattern:

* [ ] `jiff::civil::Date` <-> `DATE`
* [ ] `jiff::civil:Time` <-> `TIME`
* [ ]`jiff:civil::DateTime` <-> `TIMESTAMP`
* [x] `jiff::Timestamp` <-> `TIMESTAMPTZ`
* [ ] `jiff::SignedDuration` <-> `INTERVAL` 

Ref -

* https://github.com/launchbadge/sqlx/pull/3511
* https://github.com/sfackler/rust-postgres/pull/1164

sqlx-jiff/src/lib.rs
tisonkun: @BurntSushi Any further concerns about the wrapping style and integration style? Or shall I continue the integration for other types and adding tests.

BurntSushi: I honestly don't know. The only way this gets merged is if I have the time to dig into it and figure it whether this is the right approach overall. I don't know when I'm going to have time to do that. Hopefully soon.

BurntSushi: The problem is that I haven't gone through this exercise yet. So there's no established pattern to follow.

tisonkun: Thanks for your feedback. That's enough for the current state.

I need to evaluate whether this patch will be handled at some point so that I can continue to invest my time. As long as there is no certain blocker and we generally accept adding a `jiff-sqlx` crate, that should be fine.

BurntSushi: I don't have any certain blockers at present.

But just to be clear here, I am going to have run through and do this exercise myself, at least one or two times. I don't know how else I can definitively nail down the requirements. And there will be a lot of documentation to write as well.

I'd be willing to merge something that I know is incomplete and not polished, but I won't release it until I've had a chance to go through it myself.

tisonkun: Sure.

> I'd be willing to merge something that I know is incomplete and not polished, but I won't release it until I've had a chance to go through it myself.

This is also my flavor as a maintainer :D

tisonkun:
All bridges submitted. Note that integrations for `PgRange` and `PgTimeTz` cannot be added due to the orphan rule.

I'd suspend the task here to start `jiff-sqlx` with Postgres support first since it's my real use case and I'd avoid adding too many things at the very beginning.

tisonkun:
The failing wasm CI job doesn't seem to be relevant to this patch ..

maxcountryman:
> * `jiff::SignedDuration` <-> `INTERVAL`

I'm curious why this wouldn't be `jiff::Span`? That's how I'm doing these conversions manually.

tisonkun:
@maxcountryman 

It's possible to support multiple type mappings, so it's possible to support both Span and Duration.

However, PG's interval supports only months, days and microseconds parts. Span supports other units. I'm not sure if we can always assume 1 year=12 months, and 1 week=7days.

Anyway, we may support two types together.

maxcountryman:
>  Span supports other units. I'm not sure if we can always assume 1 year=12 months, and 1 week=7days.

As I understand, `Span` doesn't assume 12 months == 1 year (ergo keeping each unit as a discrete property). I think that might be a reason to [use `Span`](https://docs.rs/jiff/latest/jiff/struct.SignedDuration.html#when-should-i-use-signedduration-versus-span) here, right? (In particular, I believe `Span` is needed for full ISO 8601 support.)

tisonkun:
Hi @BurntSushi!

IMO code gets evolved And mature by tested in real world.

It's understandable that your context is not for such an integration, and you don't want release code that you don't look careful into.

So, I wonder if it's OK for you that I release this crate under the name `jiff-sqlx` while I'm the crate owner for all the followups. And once you or the sqlx's maintainer accept to maintain it in either upstream (for the benefits to evolve together and catch up changes in the first place), we can work a transfer solution then.

BurntSushi:
Can you please pick a different name? That way we don't need to do any transfers or have any other project expectations. 

I also thought you had this code in your own project internally? Why do you need to publish a crate to do it?

tisonkun:
> Can you please pick a different name? That way we don't need to do any transfers or have any other project expectations.

Then I'd  leave this PR here so that anyone need it can use it by copying the code. I don't think such an integration worth a dedicated name.

> I also thought you had this code in your own project internally? Why do you need to publish a crate to do it?

This is somehow like a question "why do you write opensource code". I write opensource code everyday so sharing code is by default (as shown in my GitHub profile). Also you can see that there are several other users want it. This is enough.

BurntSushi:
Yes sorry, it is totally cool to publish a crate for the good of open source. I was in a "are you yourself blocked" frame of mind versus "you want to publish a crate for others to use." So in the latter case, I think it's just a matter of either you going ahead and doing it (which is your right, although I request that you not use a `jiff-` prefix) or waiting until I'm able to dig into it.

Ecosystem integration is a big problem and I need to context switch into it. I'm just not there yet. And I'm currently thinking I'll do it with a `jiff 0.2` release, since there are a number of breaking changes I'd like to make.

maxcountryman:
My two cents: this is not the correct implementation re Postgres intervals. I think that should be addressed before publishing.

BurntSushi:
@maxcountryman I haven't looked too closely yet. It looks like only `SignedDuration` is supported here and it collapses everything down to microseconds. Can you say more about what you see as incorrect?

maxcountryman:
@BurntSushi that doesn't support ISO 8601 correct? See: https://www.postgresql.org/docs/13/datatype-datetime.html#DATATYPE-INTERVAL-OUTPUT

BurntSushi:
`SignedDuration` would still certainly emit ISO 8601.

Regardless, I'll dig into this more when I address ecosystem integration. But if you have any more explanation, I would be happy to hear it!

maxcountryman:
@BurntSushi This passage [in the docs](https://docs.rs/jiff/latest/jiff/struct.SignedDuration.html#parsing-and-printing): "Unlike the Span type, though, only uniform units are supported. This means that ISO 8601 durations with non-zero units of days or greater cannot be parsed directly into a SignedDuration" along with this from the Postgres docs, "As previously explained, PostgreSQL stores interval values as months, days, and microseconds" seem incompatible.

jiff-sqlx/src/postgres/interval.rs
```diff
@@ -0,0 +1,120 @@
...
+        2 * size_of::<i64>()
+    }
+}
+
+impl<'r> Decode<'r, Postgres> for SignedDuration {
+    fn decode(
+        value: <Postgres as Database>::ValueRef<'r>,
+    ) -> Result<Self, BoxDynError> {
+        let pg_interval = PgInterval::decode(value)?;
+        let micros = pg_interval.microseconds;
```
BurntSushi, line 67: @maxcountryman Ah so I think this is where there is a correctness issue? The non-microseconds units are being silently ignored here.

maxcountryman: This is almost certainly incorrect, yes.

However, I'm also pointing out that Postgres can be configured to output ISO 8601. Your docs say SignedDuration is not capable of parsing ISO 8601 with non-zero days. Moreover, my impression is that Spans may be preferred for ISO 8601. Regardless, Postgres will use non-zero days when configured this way so that must be handled somehow. In my own programs I just use Span.

I said as much earlier in the thread but the original author did not respond after I pointed this out.

I don't think an implementation that doesn't address this somehow should be published personally.

maxcountryman: Just to confirm there is a correctness issue here, this implementation omits two fields, [`days` and `months`](https://docs.rs/sqlx/latest/sqlx/postgres/types/struct.PgInterval.html#fields).

BurntSushi: Okay thanks, I understand now. I agree. 

I'll take a closer look later.

tisonkun: @maxcountryman Thanks for your feedback! Your analysis is correct.

The decode part is added three days ago at https://github.com/BurntSushi/jiff/pull/141/commits/2f17955624151a5327365c0d5d7bf597cf15cacd while I'd like to use SignedDuration to hold PG's interval. The background in my use case is that all the interval is written by my code and thus it's guaranteed only microseconds unit are filled. But for general purpose usage, this may not true.

There is a related discussion at https://github.com/BurntSushi/jiff/discussions/174#discussioncomment-11674374. In short, I'd first remove the decode impl for SignedDuration now because you can hardly convert days & months to seconds without a relative datetime as its design. But perhaps we can support codec over `Span`.

However, the encode part of SignedDuration should be fine.

tisonkun: Or instead, we may check months and days are 0 when decode to `SignedDuration` and return an error if so.

jiff-sqlx/src/postgres/interval.rs
tisonkun, line 1: @maxcountryman I'd re-add the impl for decoding SignedDuration.

I saw your reaction on the comment but unsure what the concrete issue you found. Maybe you can reply here.

For support of Span and convert between Span and PgInterval for sqlx codec, I'd leave it to the next PR if we get progress here, or there is a real usage requirement.

BurntSushi:
I ended up going through the exercise to do this myself. I think that was important, because it helped me learn some things I otherwise wouldn't have learned. And it also let me get confidence in the wrapper technique.

I also think that we can't do a `SignedDuration` impl here. It is way too fraught. We can actually barely even do `Span`. We can decode it just fine from PostgreSQL, but encoding one is a little trickier. I think it's possible, but it will require a little more work. I'm also somewhat worried about it because, while I think encoding is possible, it will definitely be lossy. i.e., It will lose the difference between `2 hours` and `120 minutes` (because all time units will need to collapse to microseconds).

maxcountryman:
Agree that SignedDuration should not be used. 👍

tisonkun:
> I also think that we can't do a SignedDuration impl here. It is way too fraught.

Could you elaborate a bit?

he encode part would be fine and as long as we treat illegal inputs as errors, the decode part can work on the legal partial.

tisonkun:
>  I think it's possible, but it will require a little more work. I'm also somewhat worried about it because, while I think encoding is possible, it will definitely be lossy. i.e., It will lose the difference between 2 hours and 120 minutes (because all time units will need to collapse to microseconds).

Yes. This is the background of our previous discussion at https://github.com/BurntSushi/jiff/discussions/174#discussioncomment-11673867

tisonkun:
Superseded by #240.

BurntSushi:
> > I also think that we can't do a SignedDuration impl here. It is way too fraught.
> 
> Could you elaborate a bit?
> 
> he encode part would be fine and as long as we treat illegal inputs as errors, the decode part can work on the legal partial.

Yeah I think your current implementation doesn't have footguns per se, since it does return errors and doesn't silently drop anything. And perhaps the `Encode` trait implementation is warranted at least. But I'm a little iffy on the `Decode` side since it seems surprising that it will fail in cases with non-zero months or days.

I am open to adding this back, but I'd like to see what the use cases are first. And I note that, I believe, there aren't SQLx trait implementations for either of the duration types found in `chrono` and `time`.

BurntSushi:
Oh interesting, Diesel actually has an impl for `chrono::Duration`: https://github.com/diesel-rs/diesel/blob/edd7bd32f52a52f3f451a0fd9820fc8dcf8342a8/diesel/src/pg/types/date_and_time/chrono.rs#L170-L210

It does bad juju like assuming there are always 30 days in a month.

Issue #32, opened 2024-07-22
`Span`: the `Eq`/`PartialEq` trait implementation can lead to benign-looking code that is actually buggy.


jhseu:
I'm doing a experimental migration of my project from chrono to jiff, and I've run into subtle errors involving Span comparisons. This usually involves subtracting two DateTimes and checking that its TimeDelta/Span is some value.

The documentation says:
```A Span implements the PartialEq and Eq traits, but not the PartialOrd or Ord traits. In particular, its Eq trait implementation compares for field-wise equality only. This means two spans can represent identical durations while comparing inequal```

Perhaps Span should not implement PartialEq and Eq at all. I would rather the compiler give an error if I'm doing the wrong thing rather than have a subtle bug go through. This behavior breaks the principle of least surprise. If a person unfamiliar with jiff sees `&dt1 - &dt2 == some_span` in code review, they probably would not think that it's a serious bug.

BurntSushi:
(Note: I think what you might want is [`Span::compare`](https://docs.rs/jiff/latest/jiff/struct.Span.html#method.compare).)

When doing the API design, I did think about this case and wondered what the right trade off is. In the way things are setup now, it's not that the `Eq` impl is wrong per se, because sometimes you do want to check for precise field-by-field equality. But this does bring up an important expectation mismatch with other datetime libraries that lean more heavily on absolute duration types (like `std::time::Duration` or `TimeDelta` from Chrono). In that case, doing direct equality comparisons is correct because they don't distinguish between, e.g., `120 seconds` and `2 minutes`.

This is a tough one because the `Eq` impl is incredibly convenient, especially for writing tests. Putting it into a method like `Span::fieldwise_eq` would be a major hit to ergonomics.

With all that said, the fact that `dt1 - dt2 == some_span` looks totally benign but potentially isn't is, I agree, a significant demerit to Jiff's current API.

BurntSushi:
I'm calling this a bug in the API because I think we probably need to do something about this unfortunately.

rben01:
I wonder if it would make sense to remove PartialEq from Span and then add types (available via methods or as newtype wrappers or both) that implement PartialEq in different ways. E.G., `span1.calendrical() == span2.calendrical()` for a field-wise check and `span1.temporal() == span2.temporal()` for a time delta check. This would force the user to make explicit which kind of equality they want.

BurntSushi:
[...]

PR #143, opened 2024-10-19
feat: implement Hash trait for structs that has implemented PartialEq/Eq

leiysky:
Some of the core structs(e.g. `Duration`, `Time`) has implemented `Hash` trait while some of them hasn't.

This pull request implements `Hash` trait for the rest of the structs that should have implemented it(which are `PartialEq` and `Eq`).

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -2457,6 +2457,14 @@ impl PartialOrd for Timestamp {
     }
 }
 
+impl core::hash::Hash for Timestamp {
+    #[inline]
+    fn hash<H: core::hash::Hasher>(&self, state: &mut H) {
+        self.as_second_ranged().get().hash(state);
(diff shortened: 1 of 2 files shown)


BurntSushi:
#32 probably means we're getting rid of Eq on Span, so I would hold off adding Hash to it.

leiysky:
> #32 probably means we're getting rid of Eq on Span, so I would hold off adding Hash to it.

Sure, what about the rest part?

`Timestamp` can be converted to `SignedDuration` and `Zoned` can be converted to `std::time::SystemTime`, so I think it's reasonable to implement `Hash` for them.

BurntSushi (approved):

tisonkun:
@BurntSushi may we have a new release with this patch included? I have a downstream project need it :P

PR #154, opened 2024-11-01
fmt/rfc2822: add routines explicitly for RFC 9110

### LINKED ISSUE #151: RFC 2822 should not be recommended for handling HTTP requests
According to [RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#name-date-time-formats), the preferred date/time format for HTTP is defined by [RFC 5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.3), which has replaced RFC 2822.

One notable difference is that the zone field should always use `GMT` or be omitted (implying GMT), rather than using the format `(( "+" / "-" ) 4DIGIT) / obs-zone`.

we faced [an header issue](https://github.com/astral-sh/uv/issues/8747) while using `uv`, which use jiff processing HTTP header.

BurntSushi:
It seems that RFC 9110 does not support offsets like `-0000` in the
date format, and also requires that days be padded to two digits in all
cases. This meant that Jiff's RFC 2822 output was, strictly speaking,
incompatible with a strict implementation of RFC 9110. For example, it
would output this:

    Thu, 1 Jan 1970 00:00:01 -0000

Where RFC 9110 would want this:

    Thu, 01 Jan 1970 00:00:01 GMT

I am actually still unclear as to the relationship between
RFC 2822/5322 and RFC 9110, but as far as I know, RFC 2822
has been obsoleted by RFC 5322, but RFC 5322 does not seem
to make any meaningful changes to the datetime format:
https://datatracker.ietf.org/doc/html/rfc5322#section-3.3

But, since it seems like the RFC 9110 format is a subset of the RFC
2822/5322 format, and since [MDN clearly stipulates the use of RFC
9110], we add a new method that outputs a RFC 2822 compatible timestamp
that is explicitly compatible with RFC 9110.

Fixes #151

[MDN clearly stipulates the use of RFC 9110]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since

### DIFF
--- a/src/fmt/rfc2822.rs
+++ b/src/fmt/rfc2822.rs
@@ -579,7 +579,7 @@ impl DateTimeParser {
         }
         let (day, input) = input.split_at(digits);
         let day = parse::i64(day).with_context(|| {
-            err!("failed to parse {day:?} as day", day = escape::Bytes(day),)
+            err!("failed to parse {day:?} as day", day = escape::Bytes(day))
         })?;
         let day = t::Day::try_new("day", day).context("day is not valid")?;
         let Parsed { input, .. } =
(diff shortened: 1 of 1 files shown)


BurntSushi:
No idea what's going on with the `wasm-pack` test. The error is inscrutable and it works locally. Plus, it's happening on `master`, so I'm going to go ahead and merge this.

PR #166, opened 2024-12-07
readme: one less command

tshepang:


BurntSushi (changes requested):
Can you update `lib.rs` docs too? i.e., https://docs.rs/jiff/latest/jiff/#usage

Thanks!

tshepang:
I can squash if that is preferred

BurntSushi (approved):
Thanks! I just do squash & merge for cases like this.

PR #167, opened 2024-12-09
Update chrono-tz dev-dependency from 0.9 to 0.10

musicinmybrain:
I confirmed that `cargo test --workspace` still passes.

From the [release notes](https://github.com/chronotope/chrono-tz/releases/tag/v0.10.0):

```
### Changes

- Make `OffsetName::abbreviation` return an `Option`. This reflects that numeric values such as `+11` are no longer encoded in the upstream TZDB as abbreviations (https://github.com/chronotope/chrono-tz/pull/185).
```

Since jiff and its tests don’t use this type, there is no impact and no code changes are required.


BurntSushi (approved):
Thanks!

PR #170, opened 2024-12-16
api: add `TimeZone::to_fixed_offset`

### LINKED ISSUE #169: Implement `TryFrom<TimeZone> for Offset` for fixed & utc timezones.
I would like to convert a TimeZone to an offset without access to a specific instant in time. A try_from/try_into would help me to this and handle the case where it is not possible. A other way of distinguishing fixed offsets from others non on `TimeZone` would also be fine, for example `fn to_fixed_offset(&TimeZone) -> Option<Offset>`.

Please let me know if you think this is a bad idea or not. :)

bschoenmaeckers:
closed #169

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 ============
 TBD.
 
+Enhancements:
+
+* [#169](https://github.com/BurntSushi/jiff/issues/169):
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thanks! I've rebased this on master, fixed up the docs and improved the error message a bit.

PR #177, opened 2024-12-27
Support platforms that do not have native atomics

### LINKED ISSUE #162: Support riscv32imc-unknown-none-elf
There are few uses of `Arc` and those are missing for the target.
Would it be acceptable to add `portable-atomic` and `portable-atomic-util` and the required knobs to support it?

lu-zero:
Fixes #162

e.g. for esp32c3 and such you can see it building using
```
cargo build --target=riscv32imc-unknown-none-elf --features alloc --no-default-features --features portable-atomic/unsafe-assume-single-core
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,27 @@ jobs:
     - run: cross test --verbose --target ${{ matrix.target }} --all
     - run: cross test --verbose --target ${{ matrix.target }} -p jiff-cli
 
+  # This is meant to test that Jiff's use of `Arc` from `portable-atomic-util`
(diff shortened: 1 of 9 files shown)


BurntSushi:
I've fixed this PR up a little bit:

* The `mod sync` was declared `pub(crate)` unnecessarily.
* I added some light docs.
* I added a CI job to test this.
* Most critically, I stopped enabling `alloc` unconditionally in `portable-atomic-util`. Jiff requires `alloc` opt-in, so I moved it to `alloc = ["serde?/alloc", "portable-atomic-util/alloc"]`. (I think this doesn't matter so much right now since Jiff doesn't compile in no-alloc mode, but I hope it will soon and this would have been easy to overlook!)

PR #183, opened 2024-12-31
timestamp: improve docs surrounding arithmetic panics

### LINKED ISSUE #145: Remove `Sub` implementation for `Span` to `Timestamp`
I stumbled across a little problem today. I wrote the following code:

```rust
let period = match period {
    PopularityPeriod::Daily => Span::new(),
    PopularityPeriod::Weekly => Span::new().weeks(1),
    PopularityPeriod::Monthly => Span::new().months(1),
    PopularityPeriod::Yearly => Span::new().years(1),
};

let start_period = Timestamp::now() - period;
```

And at runtime my code panicked with the following: `subtracting span from timestamp overflowed: Error { inner: ErrorInner { kind: Adhoc(operation can only be performed with units of hours or smaller, but found non-zero week units), cause: None } }`.

I find this very confusing because I don't expect add and subtract operations to be fallible (unless you have an _actual_ overflow or underflow).

Also it's not obvious from the error message _why_ substraction is limited to units of hours or smaller, instead of allowing days, weeks, months and even years.

I really think the `Sub` implementation, despite being convenient, should be removed for `Span` on `Timestamp` as I believe it to be very misleading and error-prone.

**EDIT:** Just like `Zoned` doesn't support substracting `Span` but forces to use `.checked_sub()` instead. And obviously the same goes for additions as well.

What do you think?

BurntSushi:
While this doesn't remove the arithmetic impls (I think that would be a
bridge too far), we try to make the docs a little clearer and slightly
improve the error message. The error message still doesn't say "why,"
but does hint at it a bit.

Closes #145

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -43,7 +43,7 @@ pub(crate) use err;
 /// [about error types].
 ///
 /// [about error types]: https://github.com/BurntSushi/jiff/issues/8
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Error {
     /// The internal representation of an error.
     ///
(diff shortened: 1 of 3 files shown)


mqudsi:
I'm a bit late to the jiff party and only started using it in anger this week, but I am wondering if you considering making `Sub::Output` a `Result<>` type instead (as a compromise between the current panicking behavior and removing `Sub` from `Timestamp` and co. altogether)?

Issue #147, opened 2024-10-30
feature request: week number formatting support in `strftime`


LeoniePhiline:
While looking to replace `chrono` with `jiff` in code where date formats are passed in from configuration, I realized I could not yet perform the full migration:

`jiff` does not yet provide formatting of zoned or civil date-time objects [as week number](https://docs.rs/jiff/latest/jiff/fmt/strtime/index.html#unsupported) in ISO 8601 week date (01–53, zero-padded).

`chrono` [supports this](https://docs.rs/chrono/latest/chrono/format/strftime/index.html) as `%V`, among other  week number formats (`%U`, `%W`).

`chrono::format::strftime` matches GNU coreutils `DATE(1)` (e.g. `date +%V`), which is very convenient when modernizing software.

Is there any chance `jiff` would ever change its existing `%V` (IANA time zone identifier) to a different specifier to allow for `coreutils` and `chrono` compatible specifiers?

Is the implementation of week number and week year formatting specifiers a feature that might be added to `jiff`?

BurntSushi:
The `%V` being for IANA time zone identifiers comes from `java.time`: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

I don't think that will change personally. Particularly since I see `%V` as being more important/broadly applicable than ISO 8601 week dates. Chrono itself doesn't have first class support for IANA time zone identifiers (you need to reach out to `chrono-tz` or `tzfile` for that), so it makes sense that it wouldn't have an analog.

More generally, full compatibility with other strtime APIs is pretty hard. There's no one featureful spec, and like regex engines, each implementation has their own little quirks and features with little hope for unification. I did of course try to be compatible where I could, but I think given how strtime works (a bunch of overly terse letter modifiers), there's just always going to be points where it doesn't fully line up with other implementations.

As far as supporting ISO 8601 week dates in strtime, yes, absolutely, 100% I want to do that. I just didn't do it initially because I wasn't sure about the use case. Given that `%V` is already taken, what modifier do you think Jiff should use?

LeoniePhiline:
As always, thank you for your considerate and detailed response!

> The `%V` being for IANA time zone identifiers comes from `java.time`: https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html

[...]

PR #196, opened 2025-01-10
Fix `ISOWeekDate::weekday` doc and examples

35VLG84:
As ISO-Week is Monday == day 1 based system, it would be nice to give correct example to readers in this context. 
So instead of `Weekday::to_sunday_zero_offset` use `Weekday::to_monday_one_offset` in doc and examples.

Add an assert test into the example code to highlight above.

### DIFF
--- a/src/civil/iso_week_date.rs
+++ b/src/civil/iso_week_date.rs
@@ -242,7 +242,7 @@ impl ISOWeekDate {
     /// Returns the day component of this ISO 8601 week date.
     ///
     /// One can use methods on `Weekday` such as
-    /// [`Weekday::to_sunday_zero_offset`] to convert the weekday to a number.
+    /// [`Weekday::to_monday_one_offset`] to convert the weekday to a number.
     ///
     /// # Example
     ///
@@ -253,6 +253,7 @@ impl ISOWeekDate {
     /// assert_eq!(weekdate.year(), 1948);
     /// assert_eq!(weekdate.week(), 53);
     /// assert_eq!(weekdate.weekday(), Weekday::Friday);
+    /// assert_eq!(weekdate.weekday().to_monday_one_offset(), 5);
     /// ```
     #[inline]
     pub fn weekday(self) -> Weekday {
--- a/src/civil/iso_week_date.rs
+++ b/src/civil/iso_week_date.rs
@@ -254,6 +254,7 @@ impl ISOWeekDate {
     /// assert_eq!(weekdate.week(), 53);
     /// assert_eq!(weekdate.weekday(), Weekday::Friday);
     /// assert_eq!(weekdate.weekday().to_monday_one_offset(), 5);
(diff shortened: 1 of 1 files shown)


src/civil/iso_week_date.rs
```diff
@@ -253,6 +253,7 @@ impl ISOWeekDate {
     /// assert_eq!(weekdate.year(), 1948);
     /// assert_eq!(weekdate.week(), 53);
     /// assert_eq!(weekdate.weekday(), Weekday::Friday);
+    /// assert_eq!(weekdate.weekday().to_monday_one_offset(), 5);
```
BurntSushi, line 256: Can you add `to_sunday_zero_offset` here as well?

src/civil/iso_week_date.rs
```diff
@@ -242,7 +242,7 @@ impl ISOWeekDate {
     /// Returns the day component of this ISO 8601 week date.
     ///
     /// One can use methods on `Weekday` such as
-    /// [`Weekday::to_sunday_zero_offset`] to convert the weekday to a number.
+    /// [`Weekday::to_monday_one_offset`] to convert the weekday to a number.
```
BurntSushi, line 245: I don't completely buy your reasoning for highlighting this over a Sunday zero-offset scheme _here_ personally. But I'm okay with adding both. "such as X or Y to convert the weekday to a number."

BurntSushi (changes requested):
Thanks! Some nits.

You can ignore the wasm CI failure (sigh).

src/civil/iso_week_date.rs
35VLG84: I would argue that in *the context of iso-week*,  it's actually wrong to provide Sunday-zero based example, even though the end result is the same. All documentation of iso-week is also explaining that it's scheme of Monday as day one, so it would be nice to have the same mental mapping with the example function. 

This is also slightly related to #147, as these APIs are more exposed when people are building formatters by hand.   

Lastly, on this side of the pond, Monday is more natural choice in any case :wink:

35VLG84: Sure, and the pull request is now updated.

Comment about discover-ability of this api: 

I had that mental model, that this iso-week part of the API must be bound to Zoned, as Date Line is changing based on timezone. I actually found these functions by ripgrepping the source, as I could not believe that this is not implemented. It's also confusing (to me) that Zoned have `year()`, `month()` and `day()`, but no `iso_week*`

So why not `Zoned::iso_week()` and `Zoned::iso_week_day()`? 

To me, this makes this API asymmetric and actually I don't understand why iso_week and iso_week_day are on civil side? So could it make sense change the API, or explain somewhere why they are on the civil side of things?

BurntSushi: I'd also love to hear more about why you're using the ISO week calendar in the first place.

PR #198, opened 2025-01-11
allocate exact capacity for RFC 9110 strings

zacknewman:
`fmt::rfc2822::DateTimePrinter::timestamp_to_rfc9110_string` allocates a `String` with capacity of `4`; however RFC 9110-preferred-format strings are guaranteed to have an exact length of 29. I realize this crate doesn't prioritize performance; however the fact that `String::with_capacity` was already used as opposed to `String::new` suggests this isn't an entirely inappropriate diff.

### DIFF
--- a/src/fmt/rfc2822.rs
+++ b/src/fmt/rfc2822.rs
@@ -1238,7 +1238,7 @@ impl DateTimePrinter {
         &self,
         timestamp: &Timestamp,
     ) -> Result<alloc::string::String, Error> {
-        let mut buf = alloc::string::String::with_capacity(4);
+        let mut buf = alloc::string::String::with_capacity(29);
         self.print_timestamp_rfc9110(timestamp, &mut buf)?;
         Ok(buf)
     }


BurntSushi (approved):
Yeah this looks good!

FWIW, "not prioritizing performance" doesn't mean things shouldn't be optimized or fast or that Jiff doesn't care about being fast. The key word here is _prioritization_. Prioritization is about balance and what to do when multiple things are in tension with one another. For example, the `Zoned` and `Span` types both absolutely give up performance in some cases when compared to more spartan alternatives, but they provide more convenience and make it more difficult to shoot yourself in the foot.

For something like changing the initial capacity, there's really nothing that it is in tension with. And for sure, I have spent a fair bit of time optimizing aspects of Jiff. For example, I [spent a lot of effort](https://github.com/BurntSushi/duration-unit-lookup) on an entire benchmark for one aspect of parsing the friendly duration format.

PR #199, opened 2025-01-12
tz: add support for the Android platform

### LINKED ISSUE #140: Android Platform Support
Getting the current time works as expected, but using the system Time Zone Database and obtaining the system time zone both require additional crates and lower-level Jiff methods. If Android was supported out of the box, then that would also improve ergonomics for cross-platform Android/iOS projects, since Jiff's API could be used the same way on both targets.

On Android, the Time Zone Database is in an Android-specific format. The [`android-tzdata`](https://crates.io/crates/android-tzdata) crate provides TZif data for a time zone name.

```rust
use jiff::{civil::date, tz::TimeZone};

let tz_name = "America/New_York";
let tz_data = android_tzdata::find_tz_data(tz_name)?;
let tz = TimeZone::tzif(tz_name, &tz_data)?;
let zdt = date(2023, 12, 31).at(18, 30, 0, 0).to_zoned(tz)?;
assert_eq!(zdt.to_string(), "2023-12-31T18:30:00-05:00[America/New_York]");
```

The system time zone can be obtained with the [`iana-time-zone`](https://crates.io/crates/iana-time-zone) crate.

```rust
use jiff::{tz::TimeZone, Timestamp};

let tz_name = iana_time_zone::get_timezone()?;
let tz_data = android_tzdata::find_tz_data(&tz_name)?;
let tz = TimeZone::tzif(&tz_name, &tz_data)?;
let zdt_now = Timestamp::now().to_zoned(tz);
```

BurntSushi:
Android support has two prongs:

* The special Android concatenated time zone database will now be read
  by Jiff automatically.
* The `persist.sys.timezone` Android property is read to determine the
  system's configured IANA time zone identifier.

Closes #140


src/tz/system/android.rs
```diff
@@ -0,0 +1,309 @@
...
+        // implementation is probably allocating something in order to create
+        // it... right? But there's no API to free the pointer returned. And
+        // no other indication as to its lifetime. Once again, C is awful.
+        //
+        // [1]: https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/system_properties.h#44
+
+        let mut buf = Vec::new();
+        // SAFETY: `name` is a valid NUL terminated string and
+        // `system_property_find` is a valid function read from `dlsym`
+        // according to the declaration in `include/sys/system_properties.h`.
```
fbstj, line 188: I believe this is meant to refer to `system_property_read` & `prop_info` being valid instead?

Cargo.toml
```diff
@@ -13,10 +13,13 @@ This library is heavily inspired by the Temporal project.
 categories = ["date-and-time", "no-std"]
 keywords = ["date", "time", "calendar", "zone", "duration"]
 edition = "2021"
-exclude = ["/.github", "/tmp"]
 autotests = false
 autoexamples = false
 rust-version = "1.70"
+# We include `/tests/lib.rs` to squash a `cargo package` warning that the
+# `integration` test target is being ignored. We don't include anything else
+# so tests obviously won't work, but it makes `cargo package` quiet.
```
martinvonz, line 21: Would it be a problem for the Android build for some reason if we included the entire `tests/` directory so tests work?

Context: I'm trying to import this crate in our repo at work and the tooling takes the files from the crate (not the Git commit) and tries to set up tests. I can easily configure it not to set up tests, so it's not a problem. I'm just wondering if there's a good reason for not including tests in the crate.

BurntSushi: The reason is that the test ls bloat the binary. And pretty much everyone wastes bandwidth on it. As I understand it, many crates do this, so I assume y'all already have proceses in place for dealing with this.

martinvonz: > The reason is that the test ls bloat the binary.

Do you mean the archive that's published on crates.io? Because I would assume that Rust binaries would not be bloated by tests in crates they depend on, right?

> As I understand it, many crates do this, so I assume y'all already have proceses in place for dealing with this.

I've imported hundreds of crates (~250?) into Google's monorepo. It's not uncommon that tests fail for various reasons (e.g. attempting to read files that are not in the crate), but I don't think it's common that tests don't compile. I may just have forgotten some such cases, though.

Anyway, we do have processes to not even attempt to build the tests, so no problem.

BurntSushi: Sorry, yes, I mean the archive downloaded from crates.io.

I did this when I added a test tzdata file for Android. Even compressed, it adds an extra 100KB to the artifact. It wasn't obviously worth it to me to imclude it or spend the effort to make tests work in its absence. So I figured I would just remove everything.

I recently did this for bstr as well upon request. I'm considering doing it for my other crates as well.

How important is running tests from the crate downloaded for you? Do you commonly catch regressions?

martinvonz: Thanks for explaining. That sounds reasonable.

> How important is running tests from the crate downloaded for you? Do you commonly catch regressions?

I'm not on the Rust team, but I think it's not very important. As an indication, we typically don't add build targets for the tests if it requires importing additional dev dependencies.

I think I would not even have noticed if `tests/lib.rs` had not been in the archive. https://github.com/rust-lang/cargo/issues/14290 seems to be the bug for getting a way to silence the warning and not have to include the test target.

BurntSushi: Sorry, just getting back around to this. I don't think I have anything else to add, but please let me know if this winds up being an issue. I'm open to revisiting it. Philosophically, I do like the idea of being able to run tests from the crate artifact.

PR #201, opened 2025-01-12
tz: fix termux compilation on Android

### LINKED ISSUE #200: compilation failure on Android in termux shell
Looks like this does not compile on-device from a Termux shell. There are 5 instances of this error:

```
error[E0308]: mismatched types
   --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io
-6f17d22bba15001f/jiff-0.1.22/src/tz/system/android.rs:117:38
    |
117 | ...= unsafe { dlopen(cstr("libc.so\0").as_ptr(), 0) };
    |               ------ ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `*const i8`,
 found `*const u8`
    |               |
    |               arguments to this function are incorrect
    |
    = note: expected raw pointer `*const i8`
               found raw pointer `*const u8`
note: function defined here
   --> /data/data/com.termux/files/home/.cargo/registry/src/index.crates.io
-6f17d22bba15001f/jiff-0.1.22/src/tz/system/android.rs:293:8
    |
293 |     fn dlopen(filename: *const i8, flag: i32) -> *mut c_void;
    |        ^^^^^^
```

_Originally posted by @dead10ck in https://github.com/BurntSushi/jiff/issues/140#issuecomment-2585924792_

BurntSushi:
I guess some platforms use `*const u8` for C strings while most
seemingly use `*const i8`. So insert a bunch of `.cast()` calls where
appropriate.

Ref #140

Fixes #200

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -138,6 +138,7 @@ jobs:
         - powerpc64-unknown-linux-gnu
         - s390x-unknown-linux-gnu
         - x86_64-linux-android
+        - aarch64-linux-android
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # CHANGELOG
 
+0.1.23 (TBD)
+============
+TODO
+
+Bug fixes:
+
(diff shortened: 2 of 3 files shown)


dead10ck:
I checked this out locally and I can confirm it builds now! Thank you!

PR #202, opened 2025-01-13
Restore license files that were lost in 0.1.22

musicinmybrain:
When 6926d6d84fb685ded50816fa525ee147e43aaff4 explicitly listed files to include in published crates, the license files were lost from the crates. This PR restores them by adding them to the `include` lists.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Bug fixes:
 
 * [#200](https://github.com/BurntSushi/jiff/issues/200):
 Fix compilation failure on Android in a Termux shell.
+* [#202](https://github.com/BurntSushi/jiff/pull/202):
+Re-add license files to crate artifact.
 
 
 0.1.22 (2025-01-12)
@@ -18,6 +20,10 @@ read the `persist.sys.timezone` property to determine the system's current
 time zone.
 
 See [PLATFORM] for more specific information about Android support.
+
+Note that this release also removed all non-essential files (including tests
+and test data) for the artifact uploaded to crates.io. If you need or want
+these files, please open a new issue.
 
 Enhancements:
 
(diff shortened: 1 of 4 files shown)


musicinmybrain:
(It’s a shame that I won’t be able to run tests in Fedora’s [`rust-jiff`](https://src.fedoraproject.org/rpms/rust-jiff) package anymore, since we must package from published crates, and 6926d6d84fb685ded50816fa525ee147e43aaff4 removed files needed to compile and run the tests.)

decathorpe:
> I made this change as a result of a request to remove test data in another crate: [BurntSushi/bstr#200](https://github.com/BurntSushi/bstr/issues/200)

To be fair, I caught this in bstr too, I just didn't have time to file an issue / comment on the PR yet. In general, for distribution packages (Fedora and debian, to my knowledge), we use crate sources as distributed on crates.io, and run tests (if possible). I understand that it's a trade-off between making downloads larger for *everyone* or making it possible to run tests from published artifacts - and that in most cases the balance tips towards making the downloads smaller.

BurntSushi:
@decathorpe I would like to hear more about your use case. For example, how often do y'all catch regressions as a result of running tests using the crate artifact?

musicinmybrain:
> In some cases, we can make the tests work by adding an additional source archive, like the archive from GitHub. This is [allowed by our policies](https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_package_sources) as long as the GitHub archive is provably used only for testing, and doesn’t contribute to the crate library sources we ship in the resulting `-devel` package. This tends to be a little tedious, so we are more likely to just disable tests when the crate doesn’t contain necessary files for testing, but the extra effort worth it in some cases. This might be one of those cases, and I’ll probably attempt this approach with the next release.

This approach was successful. I’m preparing to [ship `rust-jiff-0.1.23` in Fedora](https://src.fedoraproject.org/rpms/rust-jiff/pull-request/10), with the tests re-enabled, and everything looks fine.

PR #225, opened 2025-01-27
dev-dependencies: remove dependency on serde-yml

BurntSushi:
... and replace it with serde-yaml. This is non-ideal since serde-yaml
is deprecated, but I'll take that over the shenanigans going on with
serde-yml (that I was unaware of, since this was a dev-dependency).

Ref https://x.com/davidtolnay/status/1883906113428676938


BurntSushi:
This PR is in `jiff 0.1.28` on crates.io.

PR #237, opened 2025-02-02
[v0.1] crates: move all sub-crates into `crates/` directory

BurntSushi:
I should have just done this from the start. I hate doing these kinds of
moves because it fucks up revision history. But if I'm going to be
adding `jiff-icu`, `jiff-chrono`, `jiff-sqlx` and so on, then I really
think we need a sub-directory.

I'm keeping Jiff proper in `src/` though.


BurntSushi:
This is a backport of #236 to the `release/v01` branch. I should have just done the move before I merged the `v0.2` changes in #217, but I forgot.

PR #241, opened 2025-02-08
jiff-diesel: add integration crate for Diesel

### LINKED ISSUE #50: add ecosystem integration to jiff? (e.g., optional dependencies on `icu`, `diesel`, `sqlx`, etc)

BurntSushi:
This is like #240, but adds a crate with wrapper types that implements
Diesel traits. Unlike with SQLx, and unless I'm missing something,
Diesel actually exposes enough of an API to implement datetime support
for MySQL. So that's included here.

Diesel does seem to use some internal privileged APIs for its own
`chrono` and `time` integration that avoids unnecessary allocations
when parsing datetimes from SQLite values. Again, unless I'm missing
something, Jiff is forced to allocate into a `String` first. (But Jiff
only really needs a `&[u8]`.)

I found this experience, along with SQLx, to be absolutely mind-numbing.
Just writing out the example code (which I also used for ad hoc testing)
took an incredible amount of time. I spent _way_ more time playing
fucking type tetris with both SQLx and Diesel than I did anything else
_combined_. It's utterly ridiculous. This further solidifies my opinion
that when you publish crates with an obscene amount of inter-connected
traits, the resulting API becomes very difficult to use.

I'm happy to iterate on the implementation and APIs of this crate (and
`jiff-sqlx`) after an initial release. But I very much appreciate
reviews from Diesel and SQLx experts.

I'm going to say that this closes #50 since this I think this, along
with `jiff-sqlx` and `jiff-icu`, gives us a solid foundation to build
upon. We can track more specific integrations in new issues.

Closes #50


crates/jiff-diesel/src/wrappers.rs
```diff
@@ -0,0 +1,225 @@
...
+
+/// A wrapper type for [`jiff::Span`].
+///
+/// # PostgreSQL: Limited support
+///
+/// This type _only_ has a [`diesel::deserialize::FromSql`] trait implementation
+/// for PostgreSQL. The reason for this is that encoding an arbitrary
+/// `Span` into a PostgreSQL interval requires a relative datetime.
+/// Therefore, users wanting to store a `Span` will need to explicitly use a
+/// [`diesel::pg::data_types::PgInterval`] at least at encoding time.
```
weiznich, line 196: I'm not sure I really understand the problem here. Is it that you cannot know how many days a month has without having a date? Or is the problem something different?

In the first case it might be possible to just construct the `PgInterval` via [`IntervalDsl`](https://docs.diesel.rs/2.2.x/diesel/pg/expression/dsl/trait.IntervalDsl.html) and let diesel handle this problem for you. 

Otherwise maybe provide a `fn to_pg_interval(self, date: D) -> PgInterval` method then?

weiznich (commented):
Thanks for writing this :heart:

I've added a few comments about the implementation, hopefully addressing some of the issues mentioned in the code. If you like I can try following up with a PR fixing these things.

crates/jiff-diesel/src/sqlite.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+}
+
+impl Queryable<sql_types::Timestamp, Sqlite> for DateTime {
+    type Row = DateTime;
+
+    fn build(row: DateTime) -> deserialize::Result<DateTime> {
+        Ok(row)
+    }
+}
```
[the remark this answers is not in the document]
BurntSushi: Yeah I think this ended up happening in https://github.com/BurntSushi/jiff/pull/243/files thanks to @fiadliel!

crates/jiff-diesel/src/wrappers.rs
BurntSushi: Yeah I think perhaps we can do better here. And this comment is, I agree, a little confusing.

It does look like `IntervalDst` would help here to be able to encode an arbitrary `Span` as a `PgInterval`.

Otherwise, the problem I'm most concerned about here is that encoding into a `PgInterval` is necessarily lossy. A `PgInterval` only cares about months, days and microseconds. But a `Span` is a different beast. It cares about all units. It considers, for example, `120 minutes` and `2 hours`, as distinct values in memory. But if you're using a `Span` and encode it into a `PgInterval`, both `120 minutes` and `2 hours` will get encoded into the same value. The physical duration will remain the same, but the expressed units do not.

It might be that we actually want separate wrapper types. One whose goal is to integrate with `PgInterval`, even if it's lossy. And one whose goal is to preserve the `Span` exactly as-is, but does not provide native `PgInterval` support (and e.g., is stored as `TEXT`).

crates/jiff-diesel/src/sqlite.rs
```diff
@@ -0,0 +1,170 @@
...
+        value: SqliteValue<'_, '_, '_>,
+    ) -> deserialize::Result<Timestamp> {
+        // Apparently the Diesel internal implementations get to use
+        // `RawValue::parse_string`, which avoids creating a new alloc.
+        // But I guess we have to get a full alloc here? Pretty lame if
+        // you ask me. Ideally we could get a `&str` and parse straight
+        // from that. Jiff can parse from `&[u8]`, but we also use
+        // std's float parsing, which only works on `&str`.
+        let text: String =
+            FromSql::<sql_types::Timestamp, Sqlite>::from_sql(value)?;
```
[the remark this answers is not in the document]
BurntSushi: Ah yeah that makes sense. And thanks!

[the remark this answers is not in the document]
weiznich: I've opened https://github.com/diesel-rs/diesel/pull/4480 to make the relevant functions public.

crates/jiff-diesel/src/wrappers.rs
weiznich: It's even worse that just loosing the exact encoding as `PgInterval` (or better the underlying postgres type) is not roundtrip safe and that behavior even changes in different postgres versions. See https://github.com/diesel-rs/diesel/blob/master/diesel/src/pg/expression/extensions/interval_dsl.rs#L348-L351

BurntSushi: Lovely. Thanks for the heads up!

PR #243, opened 2025-02-09
jiff-diesel: Derive the `Queryable` trait for diesel wrapper types.

fiadliel:
This reduces the amount of boilerplate code needed for each wrapper type. In addition, the implementation is more general, allowing external implementations of `Backend` to use `Queryable`.

Example expansion:
```rust
    const _: () = {
        use diesel;
        use diesel::deserialize::{self, FromSql, Queryable};
        impl<__DB, __ST> Queryable<__ST, __DB> for Timestamp
        where
            __DB: diesel::backend::Backend,
            __ST: diesel::sql_types::SingleValue,
            Self: FromSql<__ST, __DB>,
        {
            type Row = Self;
            fn build(row: Self::Row) -> deserialize::Result<Self> {
                Ok(row)
            }
        }
    };
```

### DIFF
--- a/crates/jiff-diesel/src/mysql.rs
+++ b/crates/jiff-diesel/src/mysql.rs
@@ -1,5 +1,5 @@
 use diesel::{
-    deserialize::{self, FromSql, Queryable},
+    deserialize::{self, FromSql},
     mysql::{
         data_types::{MysqlTime, MysqlTimestampType},
         Mysql, MysqlValue,
@@ -12,14 +12,6 @@ use jiff::{civil, tz};
 use crate::{Date, DateTime, Time, Timestamp, ToDiesel};
 
 static UTC: tz::TimeZone = tz::TimeZone::UTC;
-
-impl Queryable<sql_types::Datetime, Mysql> for Timestamp {
-    type Row = Timestamp;
-
-    fn build(row: Timestamp) -> deserialize::Result<Timestamp> {
-        Ok(row)
-    }
-}
 
 impl ToSql<sql_types::Datetime, Mysql> for Timestamp {
     fn to_sql<'b>(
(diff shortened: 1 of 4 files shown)


BurntSushi (approved):
Sweet. I think I originally did it this way because, when I started, I assumed there was a way to use Diesel without the derive macros. But I think I later discovered that was not true and didn't understand the shortcomings of doing it by hand.

I wonder if this is why I kept seeing `Untyped` trait impl errors whenever I tried to inject date values into raw SQL. (Writing the example code for Diesel was absolutely tortuous.)

fiadliel:
> Sweet. I think I originally did it this way because, when I started, I assumed there was a way to use Diesel without the derive macros. But I think I later discovered that was not true and didn't understand the shortcomings of doing it by hand. But I never realized that I should be implementing `FromSqlRow` and not `Queryable`...

You can definitely write this without the macro, but it's mostly boilerplate unless you're doing something special. And the slightly more complicated implementation is more flexible, being parameterized over backend. Yeah, that naming of the derive isn't obvious at all; not quite sure why they named it that way!

> I wonder if this is why I kept seeing `Untyped` trait impl errors whenever I tried to inject date values into raw SQL. (Writing the example code for Diesel was absolutely tortuous.)

Ah, I think I know what you were doing here -- the `sql_query` call has some requirements that should perhaps be more strongly spelled out, on https://docs.diesel.rs/2.2.x/diesel/fn.sql_query.html
> Unlike most queries in Diesel, sql_query will deserialize its data by name, not by index. That means that you cannot deserialize into a tuple, and structs which you deserialize from this function will need to have #[derive(QueryableByName)].

So for raw SQL, you need to embed the types into a `QueryableByName` struct, e.g.

```rust
#[derive(QueryableByName)]
struct ResultRow {
  created_at: jiff_diesel::Timestamp // just using the wrapper type directly in this case
}
```

and then the struct will be filled using the corresponding names from the resultset.

BurntSushi:
Yeah I think what I was trying to do was test the decoding and encoding trait impls in the same query, like how I do in the sqlx examples.

And yeah, if there's no way to opt out of Diesel's derive macros then there's no point in `jiff-diesel` not using them I think. I'm just so used to doing trait impls by hand (e.g., for `serde`) to avoid derive macro dependencies haha.

Anyway, thank you for the explanation! (And all of the other work you're doing. HUUUUUUGE help.)

PR #245, opened 2025-02-09
jiff-diesel: Depend on postgres_backend and mysql_backend.

fiadliel:
Diesel splits the mysql implementation between the `mysql_backend` feature which implements the types/traits/etc., and the `mysql` feature which also interfaces with the mysql client library. By depending on the backend feature, we can avoid pulling in the mysql library dependency when it's not needed (e.g. using diesel-async).

Similarly for `postgres` and `postgres_backend`.


BurntSushi (approved):
Nice!

PR #247, opened 2025-02-10
jiff-diesel: Update diesel examples to use QueryableByName and sql_query.

fiadliel:
Updated examples for diesel that exercise a more usual usage of the library (when manually writing SQL), and also test the `deserialize_as` annotation for the wrapper type.

### DIFF
--- a/examples/diesel-mysql/main.rs
+++ b/examples/diesel-mysql/main.rs
@@ -28,7 +28,7 @@ fn example_datetime_roundtrip(
     conn: &mut MysqlConnection,
 ) -> anyhow::Result<()> {
     diesel::table! {
-        jiffs {
+        datetimes {
             id -> Integer, // Diesel tables require an ID column.
             ts -> Datetime,
             dt -> Timestamp,
@@ -38,19 +38,20 @@ fn example_datetime_roundtrip(
     }
 
     #[derive(Debug, PartialEq, QueryableByName)]
+    #[diesel(table_name = datetimes)]
     #[diesel(check_for_backend(diesel::mysql::Mysql))]
-    struct Jiff {
+    struct Row {
         #[diesel(deserialize_as = jiff_diesel::Timestamp)]
-        pub ts: jiff::Timestamp,
+        ts: jiff::Timestamp,
         #[diesel(deserialize_as = jiff_diesel::DateTime)]
-        pub dt: jiff::civil::DateTime,
+        dt: jiff::civil::DateTime,
         #[diesel(deserialize_as = jiff_diesel::Date)]
-        pub d: jiff::civil::Date,
+        d: jiff::civil::Date,
         #[diesel(deserialize_as = jiff_diesel::Time)]
-        pub t: jiff::civil::Time,
+        t: jiff::civil::Time,
     }
 
-    let given = Jiff {
+    let given = Row {
         ts: "1970-01-01T00:00:00Z".parse()?,
         dt: civil::date(2025, 7, 20).at(0, 0, 0, 0),
         d: civil::date(1999, 1, 8),
@@ -59,7 +60,12 @@ fn example_datetime_roundtrip(
 
     // We need to name the columns as Diesel's sql_query matches fields by name.
     let got = sql_query(
-        "select CAST(? AS DATETIME) as ts, CAST(? AS DATETIME) as dt, CAST(? AS DATE) as d, CAST(? AS TIME(6)) as t",
+        "select
+            CAST(? AS DATETIME) as ts,
+            CAST(? AS DATETIME) as dt,
+            CAST(? AS DATE) as d,
+            CAST(? AS TIME(6)) as t
+        ",
     )
     .bind::<sql_types::Datetime, _>(&given.ts.to_diesel())
     .bind::<sql_types::Timestamp, _>(&given.dt.to_diesel())
--- a/examples/diesel-mysql/main.rs
+++ b/examples/diesel-mysql/main.rs
@@ -1,6 +1,6 @@
 use diesel::{
-    connection::Connection, dsl::sql, mysql::MysqlConnection,
-    query_dsl::RunQueryDsl, select, sql_types, ExpressionMethods,
+    connection::Connection, mysql::MysqlConnection, query_dsl::RunQueryDsl,
+    sql_query, sql_types, QueryableByName,
 };
 use jiff::civil;
 use jiff_diesel::ToDiesel;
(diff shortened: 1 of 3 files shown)


examples/diesel-mysql/main.rs
```diff
@@ -18,72 +18,55 @@ fn main() -> anyhow::Result<()> {
...
-    let got: jiff_diesel::Timestamp = query.get_result(conn)?;
-    let expected =
-        "1970-01-01T00:00:00Z".parse::<jiff::Timestamp>()?.to_diesel();
-    assert_eq!(expected, got);
+/// Performs a round-trip with all of Jiff's datetime types.
+fn example_datetime_roundtrip(
+    conn: &mut MysqlConnection,
+) -> anyhow::Result<()> {
+    diesel::table! {
+        jiffs {
```
BurntSushi, line 31: Can you change this to `datetimes`?

examples/diesel-mysql/main.rs
```diff
@@ -18,72 +18,55 @@ fn main() -> anyhow::Result<()> {
...
-
-    // Encoding jiff::civil::Time
-    let time = civil::time(23, 59, 59, 999_999_000).to_diesel();
-    let query = select(
-        sql::<sql_types::Time>("CAST('23:59:59.999999' AS TIME(6))").eq(time),
-    );
-    assert!(query.get_result::<bool>(conn)?);
+    // We need to name the columns as Diesel's sql_query matches fields by name.
+    let got = sql_query(
+        "select CAST(? AS DATETIME) as ts, CAST(? AS DATETIME) as dt, CAST(? AS DATE) as d, CAST(? AS TIME(6)) as t",
```
BurntSushi, line 62:
```suggestion
        "select
            CAST(? AS DATETIME) as ts,
            CAST(? AS DATETIME) as dt,
            CAST(? AS DATE) as d,
            CAST(? AS TIME(6)) as t
        ",
```

examples/diesel-mysql/main.rs
```diff
@@ -18,72 +18,55 @@ fn main() -> anyhow::Result<()> {
...
+        }
+    }
 
-    // Decoding jiff::civil::DateTime
-    let query = select(sql::<sql_types::Timestamp>("'2025-07-20 00:00:00'"));
-    let got: jiff_diesel::DateTime = query.get_result(conn)?;
-    let expected = civil::date(2025, 7, 20).at(0, 0, 0, 0).to_diesel();
-    assert_eq!(expected, got);
+    #[derive(Debug, PartialEq, QueryableByName)]
+    #[diesel(check_for_backend(diesel::mysql::Mysql))]
```
BurntSushi, line 41: What does this do? (I see it mentioned once in the crate docs of Diesel, but I still don't understand it.)

examples/diesel-mysql/main.rs
```diff
@@ -18,72 +18,55 @@ fn main() -> anyhow::Result<()> {
...
+    struct Jiff {
+        #[diesel(deserialize_as = jiff_diesel::Timestamp)]
+        pub ts: jiff::Timestamp,
+        #[diesel(deserialize_as = jiff_diesel::DateTime)]
+        pub dt: jiff::civil::DateTime,
+        #[diesel(deserialize_as = jiff_diesel::Date)]
+        pub d: jiff::civil::Date,
+        #[diesel(deserialize_as = jiff_diesel::Time)]
+        pub t: jiff::civil::Time,
```
BurntSushi, line 50: Do these fields need to be `pub`? If not, I'd remove it.

BurntSushi (changes requested):
Thank you! This overall LGTM, but I have a few nits. I think all of my comments on the MySQL example apply to the PostgreSQL and SQLite examples as well.

examples/diesel-mysql/main.rs
fiadliel: This adds code which tests that types line up with the database schema and available `FromSql`/`ToSql` implementations, for the specified database backend.

If everything is good, nothing changes. If there was an error (which would later cause a compilation failure):

1. it exercises some of the type-driven machinery so that you get an error if your struct definition is incorrect, even before trying to use it in a query.
2. the error it generates tries to create an error span that points at the location where the mistake was made.
3. it can end up being more specific with what particular trait implementation is missing than you would get without it.

fiadliel: So think of the naming as "check... [my code against trait implementations]... for backend... Mysql"

examples/diesel-mysql/main.rs
fiadliel: Done.

examples/diesel-mysql/main.rs
fiadliel: Done.

examples/diesel-mysql/main.rs
```diff
@@ -18,72 +18,55 @@ fn main() -> anyhow::Result<()> {
...
+    }
 
-    // Decoding jiff::civil::DateTime
-    let query = select(sql::<sql_types::Timestamp>("'2025-07-20 00:00:00'"));
-    let got: jiff_diesel::DateTime = query.get_result(conn)?;
-    let expected = civil::date(2025, 7, 20).at(0, 0, 0, 0).to_diesel();
-    assert_eq!(expected, got);
+    #[derive(Debug, PartialEq, QueryableByName)]
+    #[diesel(check_for_backend(diesel::mysql::Mysql))]
+    struct Jiff {
```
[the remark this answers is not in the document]
fiadliel: Done. However note that I was using a convention for how the struct name relates to the table. As this change breaks the necessary relationship, I've added a `#[diesel(table_name = ...)]` annotation to explicitly say what table the struct refers to.

examples/diesel-mysql/main.rs
fiadliel: Done

examples/diesel-mysql/main.rs
BurntSushi: Hmmm interesting. I guess that raises the next question, which is why is that opt-in? :-) But thank you for educating me!

BurntSushi (approved):
Nice, thank you!

examples/diesel-mysql/main.rs
weiznich: It's opt-in as it is something added after the last major release and the relevant code needs to know the backend type (`diesel::mysql::Mysql`) to perform this check. There is unfortunately no way to get that type for old impls without the attribute without breaking stuff as we cannot assume that users only enabled a single backend.

Issue #246, opened 2025-02-09
Diesel custom deserialization in derives require separate wrapper for nullable types


fiadliel:
When implementing the `Queryable` trait for a structure, one can use a custom deserialization type, as described here: https://docs.rs/diesel/latest/diesel/deserialize/derive.Queryable.html

```rust
#[derive(Queryable, Selectable)]
pub struct MyTable {
  #[diesel(deserialize_as = jiff_diesel::Timestamp)]
  pub created_at: jiff::Timestamp
}
```

When implementing the `Insertable` or `AsChangeset` traits, one can use a custom serialization type, as described at https://docs.rs/diesel/latest/diesel/prelude/derive.Insertable.html and https://docs.rs/diesel/latest/diesel/prelude/derive.AsChangeset.html

```rust
#[derive(Insertable)]
pub struct MyTable {
  #[diesel(serialize_as = jiff_diesel::Timestamp)]
  pub created_at: jiff::Timestamp
}
```

However, this does not work for nullable types, i.e. this fails:

```rust
#[derive(Queryable, Selectable)]
pub struct MyTable {
    #[diesel(deserialize_as = diesel::sql_types::Nullable<jiff_diesel::Timestamp>)]
    pub deleted_at: Option<jiff::Timestamp>,
}
```

As discussed in https://github.com/diesel-rs/diesel/discussions/4425 this is due to lack of blanket implementations for `TryFrom` or `From` in the standard library.

Would it be possible to implement nullable versions of the wrapper types for this use case? I am happy to implement a PR if so.

BurntSushi:
This seems sensible! I just wasn't aware of this at all. And those examples also look really nice. They appear to make dealing with wrapper types very easy. 

Thank you for catching this!!!

BurntSushi:
Whoops, this was closed by #248.

PR #248, opened 2025-02-10
jiff-diesel: Add Nullable wrappers for Diesel types.

fiadliel:
Adds Nullable wrappers for Diesel types. This allows the use of `Option<T>` in structs used in diesel queries/updates/etc., where `T` is a supported jiff type. Leverages existing jiff wrapper implementations, so it just forwards through (de)serialization to the already written types.

Currently this PR is built on top of PR #247, will rebase after (if) that is merged.

Also, you may want to do some reorganization, documentation updates, etc. Comment here, and I'll fix, or you can update the branch. 

It does not support using these wrapper types as expressions (e.g. filter, group by, etc.). In these cases, use `Option<jiff_diesel::WrapperType>`.

Fixes issue #246.

### DIFF
--- a/crates/jiff-diesel/src/nullable.rs
+++ b/crates/jiff-diesel/src/nullable.rs
@@ -1,0 +1,322 @@
+use diesel::{
+    backend::Backend,
+    deserialize::FromSql,
+    serialize::{Output, ToSql},
+};
+
+use crate::ToDiesel;
+
+/// A wrapper type for `Option<jiff::Timestamp>`.
+///
+/// This can be used when deriving [`diesel::deserialize::Queryable`]
+/// or [`diesel::deserialize::QueryableByName`] trait implementations.
+#[derive(Clone, Copy, Debug, diesel::deserialize::FromSqlRow)]
+pub struct NullableTimestamp(Option<crate::Timestamp>);
+
+impl NullableTimestamp {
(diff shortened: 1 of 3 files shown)


crates/jiff-diesel/src/nullable.rs
```diff
@@ -0,0 +1,272 @@
...
+    ) -> diesel::deserialize::Result<Self> {
+        FromSql::from_sql(bytes).map(NullableSpan)
+    }
+
+    fn from_nullable_sql(
+        bytes: Option<<DB as Backend>::RawValue<'_>>,
+    ) -> diesel::deserialize::Result<Self> {
+        FromSql::from_nullable_sql(bytes).map(NullableSpan)
+    }
+}
```
BurntSushi, line 272: I think these all make sense to me. I do somewhat wonder if it would be nicer to put these into a sub-module named `nullable`, and then remove `Nullable` from the type names. But I think perhaps it's not worth it if we only have two copies of each type.

BurntSushi (approved):
LGTM! Once #247 is merged, I'll bring this in.

BurntSushi:
I rebased this PR on `master`. I added a blurb about the nullable types to the crate docs and also added `to_jiff` methods on all of the nullable types (to match the non-nullable wrapper types).

BurntSushi:
Thank you!!!

Issue #252, opened 2025-02-11
jiff-diesel: Full Diesel integration needs AsExpression implemented for nullable types


fiadliel:
I was looking at using the latest published version of jiff-diesel with some private code, and quickly came up against the issue of using nullable wrappers with `Insertable`/`AsChangeset`/etc.

This requires that the type implement `AsExpression`, which had been left out of the initial implementation.

Naively deriving `AsExpression` causes a compilation error:
```
rustc: conflicting implementations of trait `ToSql<Nullable<Nullable<Timestamptz>>, _>` for type `NullableTimestamp`
downstream crates may implement trait `diesel::serialize::ToSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>, _>` for type `std::option::Option<jiff::Timestamp>`
downstream crates may implement trait `diesel::serialize::ToSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>, _>` for type `nullable::NullableTimestamp`
```

While it might be possible to get around this by manually writing the implementation, it seems easier to update the other implementations. I'll add a PR for this.

Unfortunately, it probably violates semver to include this in a patch release? I'll leave it up to you as to if/how to release it.

Issue #251, opened 2025-02-11
reconsider approach to `Span` integration for PostgreSQL in `jiff-diesel` (and probably `jiff-sqlx`)


BurntSushi:
Right now, the `Span` wrapper type in `jiff-diesel` and `jiff-sqlx` only supports decoding from a [PostgreSQL interval](https://www.postgresql.org/docs/current/functions-datetime.html). All such `Span` values have, at most, non-zero months, days and/or microseconds. All other units are always zero.

Encoding is not provided currently because it has two problems with it:

1. Since a PostgreSQL interval is only a `(months, days, microseconds)` and a Jiff `Span` has many more units than that, one has to convert a `Span` into a PostgreSQL interval using only months, days and microseconds. I believe this is possible to do in a way that always preserves the exact physical duration (so long as we can assume that there are always 7 days in a week, which Jiff does indeed do). That is, this conversion step never needs to make assumptions like "there are always 30 days in a month" or "there are always 24 hours in a day." This can be done, it just requires a bit of work.
2. While in (1) we can preserve the physical duration, we cannot preserve the actual units in a Jiff `Span` when we use a PostgreSQL interval. For example, it would erase the difference between `2 hours` and `120 minutes`.

So while (1) is solvable, (2) really isn't. (2) is a concern to me because it could be potentially surprising to drop semantic information like that silently. So there is an open question for how to deal with it. Here are some choices:

1. Don't deal with it at all. This is a status quo.
2. Accept that people know what they're doing and do the encoding step to a PostgreSQL interval.
3. Provide two different `Span` wrappers. One that integrates with PostgreSQL intervals and potentially loses information, and another wrapper that uses a plain `TEXT` field with the [friendly duration format](https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html) to non-lossily encode and decode a `Span`. The downside is that you don't get native PostgreSQL interval support.

I'm partial to (3). It makes the wrappers a bit more complicated, but not significantly so IMO. We could name them, `SpanInterval` and `SpanNonLossy` or something.

Ref https://github.com/BurntSushi/jiff/pull/241#discussion_r1950393472

fiadliel:
I'd vote for option 3 here - having a lossless option is important, but some use cases will be fine with lossy.

On how to encode the lossless option, I can think of three other encodings for PostgreSQL (varying in wacky to different degrees). Probably none of these are good enough to be used instead of text, but just to enumerate some possibilities, ordered in ability/performance level:

[...]

PR #253, opened 2025-02-11
jiff-diesel: Implement AsExpression for nullable wrappers.

fiadliel:
Implements `AsExpression` for nullable wrappers, so they can be used in SQL expressions.

Fixes: #252

### DIFF
--- a/crates/jiff-diesel/src/nullable.rs
+++ b/crates/jiff-diesel/src/nullable.rs
@@ -11,6 +11,13 @@ use crate::ToDiesel;
 /// This can be used when deriving [`diesel::deserialize::Queryable`]
 /// or [`diesel::deserialize::QueryableByName`] trait implementations.
 #[derive(Clone, Copy, Debug, diesel::deserialize::FromSqlRow)]
+#[cfg_attr(
+    any(feature = "mysql", feature = "postgres", feature = "sqlite"),
+    derive(diesel::expression::AsExpression)
+)]
+#[cfg_attr(feature = "mysql", diesel(sql_type = diesel::sql_types::Nullable<diesel::sql_types::Datetime>))]
+#[cfg_attr(feature = "postgres", diesel(sql_type = diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>))]
+#[cfg_attr(feature = "sqlite", diesel(sql_type = diesel::sql_types::Nullable<diesel::sql_types::TimestamptzSqlite>))]
 pub struct NullableTimestamp(Option<crate::Timestamp>);
 
 impl NullableTimestamp {
@@ -40,9 +47,49 @@ impl From<NullableTimestamp> for Option<jiff::Timestamp> {
     }
 }
 
-impl<DB: Backend, ST> ToSql<ST, DB> for NullableTimestamp
+#[cfg(feature = "mysql")]
+impl<DB: Backend>
+    ToSql<diesel::sql_types::Nullable<diesel::sql_types::Datetime>, DB>
+    for NullableTimestamp
 where
-    Option<crate::Timestamp>: ToSql<ST, DB>,
+    Option<crate::Timestamp>:
+        ToSql<diesel::sql_types::Nullable<diesel::sql_types::Datetime>, DB>,
+{
+    fn to_sql<'b>(
+        &'b self,
+        out: &mut Output<'b, '_, DB>,
+    ) -> diesel::serialize::Result {
+        self.0.to_sql(out)
+    }
+}
+
+#[cfg(feature = "postgres")]
+impl<DB: Backend>
+    ToSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>, DB>
+    for NullableTimestamp
+where
+    Option<crate::Timestamp>:
+        ToSql<diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>, DB>,
+{
+    fn to_sql<'b>(
+        &'b self,
+        out: &mut Output<'b, '_, DB>,
+    ) -> diesel::serialize::Result {
+        self.0.to_sql(out)
+    }
+}
+
+#[cfg(feature = "sqlite")]
+impl<DB: Backend>
+    ToSql<
+        diesel::sql_types::Nullable<diesel::sql_types::TimestamptzSqlite>,
+        DB,
+    > for NullableTimestamp
+where
+    Option<crate::Timestamp>: ToSql<
+        diesel::sql_types::Nullable<diesel::sql_types::TimestamptzSqlite>,
+        DB,
+    >,
 {
     fn to_sql<'b>(
         &'b self,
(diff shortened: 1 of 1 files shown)


BurntSushi:
This is wild. I can't make heads or tails of any of this. Sigh.

Thank you for doing this. And thank you for the other PR with examples. I was going to ask for that before I saw it. :-)

Once I find time to work on #251, I'll probably bring this in and release them both as `jiff-diesel 0.2.0`.

BurntSushi (approved):

fiadliel:
> This is wild. I can't make heads or tails of any of this. Sigh. The breaking change here isn't even clear to me, but the trait impls do indeed have different `where` clauses, so I buy that it's semver breaking.

You're more than an expert by far than me, really. I just think that if I'm getting rid of a "downstream crates may implement trait" error, I must be changing how downstream trait implementations (accidentally) could affect this. I don't think we would really want such downstream implementations, but there it is.

Let me try to explain what I understand of the code here:
Given that we already have `ToSql` for a jiff wrapper, what we now want is `ToSql` for the nullable version of that wrapper. As seen in the implementation for `Option`, the non-nullable version just passes through the encoding. The nullable version instead returns `Ok(IsNull)` because... that's what it should do in that case :)

In the `FromSql` case, we also leverage the non-nullable wrapper. The `from_nullable_sql` is called where the value could be null. The trait restriction `diesel::sql_types::SqlType<IsNull = NotNull>` proves that we're not trying to implement `FromSql` for a nullable value, as `Nullable<Nullable<_>>` isn't a valid encoding.

But I wouldn't mind if @weiznich could take a look in case I made a mistake, or there's a better encoding here.

weiznich:
I had a look at this and the problem is not that you need to change the `FromSql` and `ToSql` impls here. It's "just" missing these `#[derive(AsExpression)]` derives + the annotation to generate some glue code there. Now the real problem is that 
the derive generates the following code:

```rust
#[allow(unused_imports)]
const _: () = {
    use diesel;
    use diesel::expression::AsExpression;
    use diesel::internal::derives::as_expression::Bound;
    use diesel::serialize::{self, Output, ToSql};
    use diesel::sql_types::Nullable;
    impl<'__expr>
        AsExpression<diesel::sql_types::Nullable<diesel::sql_types::Time>>
        for &'__expr NullableTime
    {
        type Expression =
            Bound<diesel::sql_types::Nullable<diesel::sql_types::Time>, Self>;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
    impl<'__expr>
        AsExpression<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
        > for &'__expr NullableTime
    {
        type Expression = Bound<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
            Self,
        >;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
    impl<'__expr, '__expr2>
        AsExpression<diesel::sql_types::Nullable<diesel::sql_types::Time>>
        for &'__expr2 &'__expr NullableTime
    {
        type Expression =
            Bound<diesel::sql_types::Nullable<diesel::sql_types::Time>, Self>;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
    impl<'__expr, '__expr2>
        AsExpression<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
        > for &'__expr2 &'__expr NullableTime
    {
        type Expression = Bound<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
            Self,
        >;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
    impl<__DB>
        diesel::serialize::ToSql<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
            __DB,
        > for NullableTime
    where
        __DB: diesel::backend::Backend,
        Self:
            ToSql<diesel::sql_types::Nullable<diesel::sql_types::Time>, __DB>,
    {
        fn to_sql<'__b>(
            &'__b self,
            out: &mut Output<'__b, '_, __DB>,
        ) -> serialize::Result {
            ToSql:: <diesel::sql_types::Nullable<diesel::sql_types::Time> ,__DB> ::to_sql(self,out)
        }
    }
    impl AsExpression<diesel::sql_types::Nullable<diesel::sql_types::Time>>
        for NullableTime
    {
        type Expression =
            Bound<diesel::sql_types::Nullable<diesel::sql_types::Time>, Self>;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
    impl
        AsExpression<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
        > for NullableTime
    {
        type Expression = Bound<
            Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,
            Self,
        >;
        fn as_expression(self) -> Self::Expression {
            Bound::new(self)
        }
    }
};
```

which notably contains that `impl<__DB>  diesel::serialize::ToSql<Nullable<diesel::sql_types::Nullable<diesel::sql_types::Time>>,   __DB,> for NullableTime`. Now that impl is clearly not helpful here as there will be no `Nullable<Nullable<Time>>` sql type in diesel. It's there for non-nullable types.

Now I do see two options to address this:

Either don't derive `AsExpression` and manually write the following three impls:

+ `impl AsExpression<Nullable<Time>> for NullableTime`
+ `impl AsExpression<Nullable<Time>> for &NullableTime`
+ `impl AsExpression<Nullable<Time>> for &&NullableTime`

We don't need the other impls for `Nullable<Nullable<Time>>` in this case. The downside is that this is a bit more code to maintain an that writing these impls require to use stuff from the `diesel::internals` module, which is not part of the stable API

The other variant is to add the derive, generate that bit unneeded additional code + adjust the `ToSql` impl only to something like:

```rust
impl<DB: Backend>
    ToSql<diesel::sql_types::Nullable<diesel::sql_types::Time>, DB>
    for NullableTime
where
    Option<crate::Time>:
        ToSql<diesel::sql_types::Nullable<diesel::sql_types::Time>, DB>,
{
    fn to_sql<'b>(
        &'b self,
        out: &mut Output<'b, '_, DB>,
    ) -> diesel::serialize::Result {
        self.0.to_sql(out)
    }
}
```

this essentially only swaps out the generic `ST` type to the actual type that is used. This makes the impl a bit more restrictive and in case of `NullableTimestamp` it requires duplicating the impl with different types for the different backends. On the other hand this wouldn't require using an internal API. I personally wouldn't consider changing the bounds here as breaking change, because this was effectively already expressed by the `Option<crate::Time>: ToSql<ST, DB>` bound before. That bound only allowed `ST` to be exactly that type.

I think I personally would go for the second variant.

BurntSushi:
> I think I personally would go for the second variant.

SGTM. Thank you for chiming in!

fiadliel:
@weiznich Thanks a lot for your input! I know you're busy, so I appreciate it.

@BurntSushi I've updated this PR following the recommendation above.

BurntSushi (approved):
Nice, thank you!

PR #254, opened 2025-02-11
jiff-diesel: Update examples to use Insertable and Selectable derivations.

fiadliel:
This moves the majority of examples to use the `Insertable` and `Selectable` derivations from the query struct. This follows idiomatic Diesel usage. It also tests actual database round-tripping via database stores, includes testing of the `AsExpression` implementations.

The downside of this is that tables need to exist; I have created temporary tables in this PR for this purpose (which are dropped when the session is closed).

It also moves the database schema definition to a separate module, to match standard usage in diesel.

These tests require that #253 is merged, otherwise it's impossible to derive `Insertable` for nullable jiff types.

### DIFF
--- a/examples/diesel-mysql/main.rs
+++ b/examples/diesel-mysql/main.rs
@@ -1,9 +1,31 @@
 use diesel::{
-    connection::Connection, mysql::MysqlConnection, query_dsl::RunQueryDsl,
-    sql_query, sql_types, QueryableByName,
+    connection::Connection, mysql::MysqlConnection, sql_query, sql_types,
+    QueryableByName, RunQueryDsl,
 };
 use jiff::civil;
 use jiff_diesel::ToDiesel;
+
+mod schema {
+    diesel::table! {
+        datetimes {
+            id -> Integer, // Diesel tables require an ID column.
+            ts -> Datetime,
+            dt -> Timestamp,
+            d -> Date,
+            t -> Time,
+        }
+    }
+
+    diesel::table! {
+        nullable_datetimes {
+            id -> Integer, // Diesel tables require an ID column.
+            ts -> Nullable<Datetime>,
+            dt -> Nullable<Timestamp>,
+            d -> Nullable<Date>,
+            t -> Nullable<Time>,
+        }
+    }
+}
 
 fn main() -> anyhow::Result<()> {
     // To set this up, I ran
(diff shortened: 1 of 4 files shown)


examples/diesel-sqlite/main.rs
```diff
@@ -6,12 +6,60 @@ use diesel::{
...
+            d -> Nullable<Date>,
+            t -> Nullable<Time>,
+        }
+    }
+}
+
 fn main() -> anyhow::Result<()> {
     let tmpfile = tempfile::NamedTempFile::new()?;
     let path = tmpfile.path().to_str().context("invalid temporary file")?;
     let mut conn = SqliteConnection::establish(path)?;
```
weiznich, line 34:
```suggestion
    let tmpfile = tempfile::NamedTempFile::new()?;
    let path = tmpfile.path().to_str().context("invalid temporary file")?;
    let mut conn = SqliteConnection::establish(path)?;
```

I would like just use a in memory database here instead. So just `SqliteConnection::establish(":memory:")?;`

examples/diesel-mysql/main.rs
```diff
@@ -1,10 +1,32 @@
...
+    connection::Connection, mysql::MysqlConnection, sql_query, sql_types,
+    QueryableByName, RunQueryDsl,
 };
 use jiff::civil;
 use jiff_diesel::ToDiesel;
 
+mod schema {
+    diesel::table! {
+        datetimes {
+            id -> Integer, // Diesel tables require an ID column.
```
weiznich, line 11: Technically you can go without an `id` column, you just need to declare a random column to be the "primary key" via `datetimes(t) { …}` (turns the `t` column in your fake primary key). That works as long as you write these `table!` calls manually and don't relay on the fact that the primary key is actually a primary key.

weiznich (commented):
There is another way to test roundtrips in diesel without requiring tables at all: You can pass in the value as bind parameter to a select call. That results in code like this:

`diesel::select(your_value.into_sql::<Timestamp>()).get_result::<jiff_diesel::Timestamp>(conn)` which performs the same roundtrip in terms of serialization and deserialization than inserting + loading from a table.

BurntSushi:
That's a good question. I do use them to test. But I guess ideally they would serve as examples too. I guess keep them as-is. :-)

BurntSushi:
@fiadliel With #253 merged, should this get rebased?

fiadliel:
@BurntSushi It shouldn't matter, but rebased. Can merge if it looks good to you.

BurntSushi (approved):
LGTM. Thank you!

Issue #258, opened 2025-02-11
Document how `jiff-tzdb` data is generated


robertbastian:
`concatenated-zoneinfo.dat` seems to include copied data for each alias, instead of pointing aliases at the data for the canonical zone. For example, `Europe/Kiev` and `Europe/Kyiv` point at different parts of the blob, even though `Europe/Kiev` is an alias in TZDB, so the data in both is the same.

There are 597 identifiers, but only 339 non-alias zones in 2025a, so 45% of the data is probably duplicated.

It would also be useful to document which TZDB options are used when generating this file, e.g. backzone for including backzone zones, rearguard for normalizing DSTs to be positive (Ireland), etc.

BurntSushi:
This is how `zic` is invoked to build tzdb: https://github.com/BurntSushi/jiff/blob/e094b01adfc0faaeb78f2cf0b4888b028c686bee/crates/jiff-cli/cmd/generate/zoneinfo.rs#L124-L137

And this is where the resulting zoneinfo is traversed and embedded into the `dat` file: https://github.com/BurntSushi/jiff/blob/e094b01adfc0faaeb78f2cf0b4888b028c686bee/crates/jiff-cli/cmd/generate/tzdb.rs#L66

I believe it's all pretty dumb at the moment and doesn't try to do anything clever with aliases and what not.

robertbastian:
I've discovered an issue around Ireland's weird DST. The tzdb says that Ireland does negative DST in the winter, and normal time in the summer, *unless* you build it with the `rearguard` feature, in which case it normalizes it to standard in the winter, positive DST in the summer.

Now, Jiff doesn't have any control over whether the system tzdb uses rearguard or not, and I don't know if this is even observable in the tzifs (other than looking for the specific Ireland issue). The issue this introduces is that the `Dst` flag for Ireland will differ between systems. My system tzdb (Mac) uses rearguard, however `jiff-tzdb` does not use rearguard, so there's a behaviour difference even if both are on the same tzdb version.

ICU4X requires the `ZoneVariant` to use the rearguard semantics, i.e. DSTs are positive. It would be nice if jiff could correct for that.

BurntSushi:
OK, so I wanted to try and demonstrate the problem in terms of Jiff behavior. Here's a Rust program, with the `tzdb-bundle-always` feature enabled:

```rust
use jiff::{tz::TimeZoneDatabase, Timestamp};

fn main() -> anyhow::Result<()> {
[...]

PR #259, opened 2025-02-11
Deduplicate tzif data in `jiff-tzdb`

robertbastian:
This reduces the blob size from 336k to 200k.

You might want to regenerate this on your machine, I get the following error even on master:

```
failed to parse TZif data from America/Ojinaga: expected last transition to have DST offset of -06, but got -05 according to POSIX TZ string CST6CDT,M3.2.0,M11.1.0
```

I commented out the validation in order to generate the file.

#258


BurntSushi:
Thanks! This LGTM. I was able to re-generate the data without any errors. Not sure what's going on there with the error you got. It's coming from an inconsistency detected between the POSIX time zone and the last transition in the TZif data.

(I generate `jiff-tzdb` straight from the IANA tzdb releases, not from `/usr/share/zoneinfo`. Not sure if that's related or not.)

robertbastian:
It's probably some `zic` platform weirdness, given that there's a binary diff as well.

BurntSushi:
This ended up shrinking the size of `jiff-tzdb` on crates.io (which is probably compressed) from 80KB to 60KB. Nice!

PR #260, opened 2025-02-12
Add crate for integration with the AWS SDK for Rust

snaggen:
I have implemented some basic conversion from the AWS Smithy type `DateTime`  into/from `Timestamp` and `Zoned`.

### DIFF
--- a/crates/jiff-aws/README.md
+++ b/crates/jiff-aws/README.md
@@ -1,0 +1,6 @@
+jiff-aws
+=========
+This crate provides traits and methods to [`jiff`] that implement some basic
+conversion methods to/from the datetime types used in 
+the [AWS SDK for Rust](https://aws.amazon.com/sdk-for-rust/).
+
--- a/crates/jiff-aws/src/lib.rs
+++ b/crates/jiff-aws/src/lib.rs
@@ -1,0 +1,163 @@
+/*!
+This crate provides type convertion for [Jiff](jiff) and
+Aws Smithy Type [DateTime] used
+in the [AWS SDK for Rust](https://aws.amazon.com/sdk-for-rust/).
+
+```
+use jiff_aws::ConvertAwsDateTime;
+use jiff_aws::ConvertJiffTypes;
+use jiff::Timestamp;
+use jiff::Zoned;
+use aws_smithy_types::DateTime;
+
+let datetime = DateTime::from_secs_and_nanos(1627680004,123000000);
+let timestamp: Timestamp = datetime.try_into_timestamp().unwrap();
+assert_eq!( "2021-07-30T21:20:04.123Z".parse::<Timestamp>().unwrap(), timestamp);
+
+let timestamp = Timestamp::try_from_aws(datetime).unwrap();
+assert_eq!( "2021-07-30T21:20:04.123Z".parse::<Timestamp>().unwrap(), timestamp);
+
+let zoned = datetime.try_into_zoned().unwrap();
+assert_eq!( "2021-07-30T21:20:04.123+00:00[UTC]".parse::<Zoned>().unwrap(), zoned);
+
+let zoned = Zoned::try_from_aws(datetime).unwrap();
+assert_eq!( "2021-07-30T21:20:04.123+00:00[UTC]".parse::<Zoned>().unwrap(), zoned);
+
+let other: DateTime = timestamp.into_aws_datetime();
+assert_eq!(datetime, other);
+
+let other = DateTime::from_timestamp(timestamp);
+assert_eq!(datetime, other);
+
+let other: DateTime = zoned.clone().into_aws_datetime();
+assert_eq!(datetime, other);
+
+let other = DateTime::from_zoned(zoned);
+assert_eq!(datetime, other);
+
+```
+*/
+
+use aws_smithy_types::DateTime;
+use jiff::{tz::TimeZone, Timestamp, Zoned};
+pub use traits::{ConvertAwsDateTime, ConvertJiffTypes};
+mod traits;
+
+impl ConvertAwsDateTime for Timestamp {
+    type Error = jiff::Error;
+
+    fn into_aws_datetime(self) -> DateTime {
+        DateTime::from_timestamp(self)
+    }
+
+    fn try_from_aws(value: DateTime) -> Result<Self, Self::Error> {
+        jiff::Timestamp::new(value.secs(), value.subsec_nanos() as i32)
+    }
+}
+
+impl ConvertAwsDateTime for Zoned {
+    type Error = jiff::Error;
+
+    fn into_aws_datetime(self) -> DateTime {
+        DateTime::from_zoned(self)
+    }
+
+    fn try_from_aws(value: DateTime) -> Result<Self, Self::Error> {
+        let timestamp =
+            jiff::Timestamp::new(value.secs(), value.subsec_nanos() as i32)?;
+        Ok(timestamp.to_zoned(TimeZone::UTC))
+    }
+}
+
+impl ConvertJiffTypes for DateTime {
+    type Error = jiff::Error;
+
+    fn from_timestamp(timestamp: Timestamp) -> Self {
+        let (seconds, nanos) = if timestamp.subsec_nanosecond() < 0 {
+            (
+                timestamp.as_second() - 1,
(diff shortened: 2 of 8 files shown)


robertbastian:
Wouldn't this be better in [aws_smithy_types_convert](https://docs.rs/aws-smithy-types-convert/0.60.8/aws_smithy_types_convert/date_time/trait.DateTimeExt.html) like the `chrono` and `time` conversions?

snaggen:
I created a PR for that previously, but they didn't want to add it since it isn't stable yet.

BurntSushi:
Ref https://github.com/smithy-lang/smithy-rs/pull/3826

I'll get to reviewing this soon, after I get through the batch of changes related to core-only work.

snaggen:
> Ref [smithy-lang/smithy-rs#3826](https://github.com/smithy-lang/smithy-rs/pull/3826)
> 
> I'll get to reviewing this soon, after I get through the batch of changes related to core-only work.

No, worries.... it is just some simple convenience helpers, so it is not like my code is dependent on this :)

crates/jiff-aws/README.md
```diff
@@ -0,0 +1,11 @@
...
+=========
+This crate provides wrapper types for [`jiff`] that implement the necessary
+traits for integration with [`sqlx`].
+
+[`diesel`]: https://docs.rs/sqlx
+[`jiff`]: https://docs.rs/jiff/0.2
+
+### Documentation
+
+https://docs.rs/jiff-sqlx
```
stormshield-gt, line 11: I think you forgot to replace `sqlx` with `aws`

snaggen: Thanks for the review, yes this is an oversight on my side... I will fix

snaggen: I have updated the README now

crates/jiff-aws/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+impl ConvertAwsDateTime for Timestamp {
+    type Error = jiff::Error;
+
+    fn into_aws_datetime(self) -> DateTime {
+        DateTime::from_timestamp(self)
+    }
+
+    fn try_from_aws(value: DateTime) -> Result<Self, Self::Error> {
+        jiff::Timestamp::new(value.secs(), value.subsec_nanos() as i32)
```
fiadliel, line 54: Smithy times always have a positive value for nanos, and the value always counts forward in time: https://github.com/smithy-lang/smithy-rs/blob/b34250611c130cb71715ae68e8c6fae86c72677f/rust-runtime/aws-smithy-types/src/date_time/mod.rs#L59

However in jiff, the microsecond value may count either forward or backwards in time, and may be negative: https://github.com/BurntSushi/jiff/blob/460a0bc039f1a8c8381c97d4d207ff755157bee1/src/timestamp.rs#L701

So, I think this will give a wrong conversion for timestamps before the Unix epoch.

fiadliel:
The nanosecond portion of a jiff Timestamp are negative for times before the Unix epoch (it ranges between +1s and -1s), this should be taken into account (e.g. the type casts between u32 and i32 are incorrect).

snaggen:
> The nanosecond portion of a jiff Timestamp are negative for times before the Unix epoch (it ranges between +1s and -1s), this should be taken into account (e.g. the type casts between u32 and i32 are incorrect).

The smithy datetime is defined as seconds and nano seconds since unix epoch, so a simple check for that should fix this. I will have a look at it later.

snaggen:
@fiadliel  Ok, it seems that I was wrong about DateTime being only positive. But I have added nano second conversions for times before epoc. Please have a look and see if it looks correct to you. I checked for conversion the other way also, but that seems to be handled by jiff automatically.

crates/jiff-aws/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    fn from_timestamp(timestamp: Timestamp) -> Self {
+        let nanos = if timestamp.subsec_nanosecond() < 0 {
+            (1_000_000_000 + timestamp.subsec_nanosecond()) as u32
+        } else {
+            timestamp.subsec_nanosecond() as u32
+        };
+
+        DateTime::from_secs_and_nanos(timestamp.as_second(), nanos)
+    }
```
fiadliel, line 83: Take, for example, a time -1ns before the epoch.

In jiff, this will be represented by `{seconds: 0, nanos: -1)`.
In smithy, this will be represented by `{seconds: -1, nanos: 999_999_999}`.

So both the nanosecond and second components are affected.

Some round-tripping unit tests might be useful to verify some of the expected properties (e.g. converting to smithy and back again should result in the same time).

snaggen:
Added some tests and ensured the calculations are correct for negative epoch

crates/jiff-aws/src/lib.rs
snaggen: I have fixed the calculations and added tests.

fiadliel:
I think that fixes the calculation around times before 1970. Thanks!

One last thing: I think you should reconsider being able to convert `DateTime` into `Zoned` (or remove `Zoned` altogether), ref. the opening comment in this PR: https://github.com/BurntSushi/jiff/pull/240 where `Zoned` support was left out of the `jiff-sqlx` crate.
But look, I'm just a drive-by commenter, so it's not an official request :)

snaggen:
I don't really have any strong opinions about including Zoned or not. But I added it since this is to make things a bit more convenient, not having to jump through a Timestamp first.  And it is clearly documented that it creates a Zoned for timezone UTC, so I don't really see any issues with having it included.

snaggen:
@BurntSushi have you had time to look at this?

BurntSushi:
Thanks for the ping.

In terms of whether I'm willing to bring this crate on and maintain it indefinitely, I'm somewhat unsure. It does look pretty lightweight. On the other hand, the [smithy-lang people seem on board with supporting Jiff once it's 1.0](https://github.com/smithy-lang/smithy-rs/pull/3826), and that shouldn't be too far away.

The other consideration here is the complexity of the actual conversion. For `icu`, for example, the conversions are not just instants in time, but civil times as well, and those have more failure points. But here, the only relevant target type is [`aws_smithy_types::date_time::DateTime`](https://docs.rs/aws-smithy-types/latest/aws_smithy_types/date_time/struct.DateTime.html), which is an _instant_ in time. Which means I don't think an entire crate for this is really buying us much.

How would you use a crate like this? Is this something where you could define a helper function on your own and just use that? I also don't have a good sense of how many people would actually benefit from this, which makes it hard to determine if this is worth maintaining indefinitely on my end.

Separately from the question of whether we _should_ do this, the actual change here would need some work. Specifically:

* It should follow the API design pattern established in `jiff-icu`.
* We shouldn't have conversions to `Zoned`. A conversion that automatically uses UTC isn't buying much, and that sort of thing really should be determined by the caller. I would in theory be okay with adding this conversion in the future if there was strong evidence that it was very common and folks were stumbling on it though.

snaggen:
Fair points! I'll wait for 1.0 and try to add it to the smithy-lang crate

And about usage, sure this could easily be a helper function you define by your self. However, it is always convenient to get something that is tested, especially when there are non-obvious cases like before epoch where you have to recalculate things a little bit.

BurntSushi:
Another point of clarity here is that I'm definitely not trying to be a receptacle of Jiff integrations with other crates. I did SQLx and Diesel because I perceived folks were bouncing off of the lack of Jiff integration for those specific crates _and_ because they were unwilling to add upstream support, and I did `icu` because I plan for `icu` to be a paved path toward non-Gregorian calendar support and localized formatting. The story for SQLx and Diesel integration is a fair bit more compelling, because if you don't have a Jiff integration, you end up needing to write a fair amount of arcane boiler plate to get it to plug into the SQLx/Diesel machinery. It's very much non-trivial.

> Fair points! I'll wait for 1.0 and try to add it to the smithy-lang crate

All righty, thanks. I'll close this out then. If this doesn't end up working, please feel free to swing back around here and we can re-evaluate.

And I do acknowledge that having a tested conversion routine is beneficial. For me, it's more about whether there's enough people hitting this that it's worth me maintaining in order to overcome what I hope is a short-term gap. If, once Jiff is 1.0 and the smithy folks still won't add support for it (and it seems like they would, or at least, that's I think what they telegraphed), then I'd be more open to supporting this longer term.

PR #268, opened 2025-02-17
doc: beef up docs for methods that set sub-second units

### LINKED ISSUE #261: Suprising behaviour: setting `.with().nanosecond(0)` doesn't set the whole subsecond part to 0, must use `.subsec_nanosecond()`
Hi,

This is not an issue per se, but just a surprising (and non-intuitive to me) behavior that led to a bug in one of my projects. I was expecting that `.with().nanosecond(0)` on a `Zoned` will reset its whole sub-second part to `0` (akin to the `chrono`'s `.with_nanosecond()`). Turns out it only resets the nano part of the sub-second. I.e. `0.123456789` will become `0.123456`. For the expected behavior one should use `with().subsec_nanosecond()`.

Knowing that, I checked the docs more carefully, and I only found mention of this in the `DateTimeWith` method descriptions (relationship between `subsec_nanosecond()` and separate sub-second methods). But it never clearly mentions that setting the sub-second part it will only set that part of the sub-second without resetting other parts.

I think it should be addressed, maybe with more prominent documentation mention or to make it so that `.with().microsecond(100)` changes `0.123456789` to `0.000100000` instead of `0.123100789`.

Thank you for you work on the crate!

BurntSushi:
For example, there is `Time::nanosecond` and `Time::subsec_nanosecond`.
The former sets the individual nanosecond component and does not change
the millisecond or microsecond components. But the latter sets the
entire fractional second component to nanosecond precision.

Fixes #261

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # CHANGELOG
 
+0.2.2 (TBD)
+===========
+TODO
+
+Bug fixes:
+
+* [#261](https://github.com/BurntSushi/jiff/issues/261):
+Improve the documentation for `ZonedWith::nanosecond` and
+`ZonedWith::subsec_nanosecond`.
+
+
 0.2.1 (2025-02-16)
 ==================
 This release includes a massive number of optimizations that significantly
--- a/src/civil/datetime.rs
+++ b/src/civil/datetime.rs
@@ -4183,6 +4183,11 @@ impl DateTimeWith {
     ///
     /// This overrides any previous millisecond settings.
     ///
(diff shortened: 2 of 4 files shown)


curiousdannii:
Can the docs for the getting functions also be updated? I was going to flag that the docs for `Zoned.microsecond()` were wrong and should say `0..=999_999` instead, but now I've realised that they're referring only to the microsecond "component", not the whole subsecond value rounded to microseconds.

BurntSushi:
I guess that's probably wise, given the confusability here. Idk when I'll get to it, but PRs are welcome.

PR #280, opened 2025-02-27
ci: try to speed up CI

BurntSushi:
The jobs that run `./test`, which test a bunch of different feature
combinations, take quite some time. I think a lot of that time is just
spent compiling the different feature combinations. But I suspect we are
also spending a lot of time running doc tests.

So this switches to Rust 2024 *just* for the jobs that run `./test`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,15 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
         toolchain: ${{ matrix.rust }}
+    # In Rust 2024, doc tests are (mostly) all compiled into a single
+    # executable and then run once. This is a lot faster for Jiff, which
+    # has 1000+ doc tests. Since this job is where we run most of our
+    # tests (and we run MSRV on a different job), we just bump to Rust 2024
+    # to take advantage of this.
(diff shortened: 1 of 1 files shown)


BurntSushi:
I think the other looming problem here is that Jiff's dev-dependencies are kind of ridiculous. And all of those need to get re-compiled for each feature combination, even though they aren't actually always used! (Since we do a lot of explicit `--lib` and `--test integration` for running tests.)

There's probably a few ways to go about addressing that... Hmmmm. Part of the problem is that some non-trivial subset of them are used in doc tests inside of `jiff`. But I think many of them come from the `_documentation` sub-module. Maybe we can arrange things differently. Of course, I think we could add internal dev-only crate features for this. But that's kind of a bummer?

BurntSushi:
OK, so `test-stable` took 8m, `test-beta` took 15m, `test-nightly` took 10m and `test-macos` took 7m. Looking at #278 as the last PR before this change, it took 9m, 15m, 13m and 6m, respectively. So that... kinda looks like an improvement? But the variation here is just wild.

I'll run with this for now, but if it looks like it doesn't consistently help, I'll take it back out. Kind of a bummer, but it might mean there is something I don't understand about why `./test` is so slow.

PR #282, opened 2025-02-28
Implement `Sum` for `SizedDuration`

### LINKED ISSUE #274: `impl std::iter::Sum for SignedDuration`
I have a usecase where I want to sum up some durations, that's possible via `.reduce(|a, b| a + b)` but it would be nice to be able to use `.sum()` too.

daalfox:
Closes #274

### DIFF
--- a/src/signed_duration.rs
+++ b/src/signed_duration.rs
@@ -2521,7 +2521,6 @@ fn parse_iso_or_friendly(bytes: &[u8]) -> Result<SignedDuration, Error> {
 
 #[cfg(test)]
 mod tests {
-    use core::i64;
     use std::io::Cursor;
 
     use alloc::string::ToString;
--- a/src/signed_duration.rs
+++ b/src/signed_duration.rs
@@ -2166,6 +2166,24 @@ impl core::ops::Mul<i32> for SignedDuration {
     }
 }
 
+impl core::iter::Sum for SignedDuration {
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        iter.fold(Self::new(0, 0), |acc, d| acc + d)
+    }
+}
+
+impl<'a> core::iter::Sum<&'a mut Self> for SignedDuration {
+    fn sum<I: Iterator<Item = &'a mut Self>>(iter: I) -> Self {
+        iter.fold(Self::new(0, 0), |acc, d| acc + *d)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/signed_duration.rs
```diff
@@ -2166,6 +2166,24 @@ impl core::ops::Mul<i32> for SignedDuration {
...
+        iter.fold(Self::new(0, 0), |acc, d| acc + *d)
+    }
+}
+
+impl<'a> core::iter::Sum<&'a Self> for SignedDuration {
+    fn sum<I: Iterator<Item = &'a Self>>(iter: I) -> Self {
+        iter.fold(Self::new(0, 0), |acc, d| acc + *d)
+    }
+}
+
```
BurntSushi, line 2186: Hmmm, std only has the `T` and `&T` impls for primitive integers. Could you drop the `&mut T` impl here? It's probably harmless, but I'd rather match what std does for Chesterton fence reasons until there's a concrete reason to do otherwise.

daalfox:
Not sure if it's currect but I implemented `Sum` for `Iterator<Self>`, `Iterator<&Self>` and `Iterator<&mut self>`. Felt like all of them should be possible for the users.

BurntSushi (changes requested):
Thanks! One small nit.

src/signed_duration.rs
daalfox: Done

src/signed_duration.rs
```diff
@@ -2509,6 +2521,7 @@ fn parse_iso_or_friendly(bytes: &[u8]) -> Result<SignedDuration, Error> {
 
 #[cfg(test)]
 mod tests {
+    use core::i64;
```
BurntSushi, line 2524: Sorry, missed this on the first go-round. But you should be able to remove this import entirely. [`i64::MAX`](https://doc.rust-lang.org/std/primitive.i64.html#associatedconstant.MAX) is defined on the `i64` type itself.

(In the olden days, you _did_ need this import since `MAX` on the `i64` type didn't exist. So instead, you had to use the `MAX` constant in the `core::i64` module. But you've been able to elide this import since Rust 1.43, and Jiff's MSRV is Rust 1.70.)

BurntSushi (changes requested):

src/signed_duration.rs
daalfox: yep

PR #289, opened 2025-03-06
zoned: improve docs for `Zoned::subsec_nanosecond`

### LINKED ISSUE #283: Add subsec_millisecond and subsec_microsecond to Zoned?
`Timestamp` has `subsec_millisecond`, `subsec_microsecond`, and `subsec_nanosecond`, but `Zoned` only has `subsec_nanosecond`. For consistency could `Zoned` have the others?

BurntSushi:
It is tempting to think of this method as just being a shortcut for
`zdt.timestamp().subsec_nanosecond()`, but it actually isn't! It's
returning the fractional seconds on the *civil* datetime, not the
timestamp. These are usually the same for times after the Unix epoch,
but can differ for times before it.

While the original request in #283 asks for `subsec_millisecond()` and
`subsec_microsecond()` on `Zoned` in order to be consistent with
`Timestamp`, I'm going to pass on those for now. In particular, since
these would return the fractional second value from the *civil*
datetime, `subsec_millisecond()` would always be equivalent to
`millisecond()`. `subsec_microsecond()` wouldn't be the same as
`microsecond()` (just like `subsec_nanosecond()` isn't the same as
`nanosecond()`), but I find that this just overall adds to the confusion
of the methods here. And if you do need `subsec_microsecond()`, you can
just do `subsec_nanosecond() / 1_000`.

The reason that these additional methods make sense for `Timestamp` is
that `Timestamp` doesn't have a civil datetime. So there are no
individual `millisecond()` or `microsecond()` units. A `Timestamp` is
closer to a `SignedDuration` than a `civil::DateTime`.

Closes #283

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -262,6 +262,16 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
         toolchain: stable
+    # In Rust 2024, doc tests are (mostly) all compiled into a single
+    # executable and then run once. This is a lot faster for Jiff, which has
+    # 1000+ doc tests. Since we run doc tests in this job, we just bump to Rust
+    # 2024 to take advantage of this.
+    - name: Switch to Rust 2024
+      shell: bash
+      run: sed -i.bak 's/^edition = "2021"$/edition = "2024"/' Cargo.toml
+    - name: Bump rust-version to Rust 1.85
+      shell: bash
+      run: sed -i.bak 's/^rust-version = "1\.70"$/rust-version = "1.85"/' Cargo.toml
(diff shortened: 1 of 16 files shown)


BurntSushi:
Woohoo, okay, I've got CI times now down to ~6.5 minutes or so. They were at 17 minutes. See the commits for more details, but basically, I split apart the behemoth `./test` script in multiple pieces.

There also seems to be something funny going on with the core-only tests and `insta`. They are _quite_ slow. Since I was force passing them anyway (because I couldn't get `insta` to work in that context), I've just `cfg`'d them out for now. And that has sped up those tests substantially.

PR #303, opened 2025-03-30
span: avoid cloning when rounding

Patryk27:
Quick fix - there's no need to actually clone the underlying dates, since we can simply borrow them.

### DIFF
--- a/src/util/borrow.rs
+++ b/src/util/borrow.rs
@@ -14,6 +14,15 @@ pub(crate) enum DumbCow<'a, T> {
     Borrowed(&'a T),
 }
 
+impl<'a, T> DumbCow<'a, T> {
+    pub(crate) fn borrowed(&self) -> DumbCow<'_, T> {
+        match *self {
+            DumbCow::Owned(ref this) => DumbCow::Borrowed(this),
+            DumbCow::Borrowed(ref this) => DumbCow::Borrowed(this),
+        }
+    }
+}
+
 impl<'a, T> core::ops::Deref for DumbCow<'a, T> {
     type Target = T;
     fn deref(&self) -> &T {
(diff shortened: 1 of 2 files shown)


src/util/borrow.rs
```diff
@@ -14,6 +14,15 @@ pub(crate) enum DumbCow<'a, T> {
     Borrowed(&'a T),
 }
 
+impl<'a, T> DumbCow<'a, T> {
+    pub fn borrowed(&self) -> DumbCow<T> {
```
BurntSushi, line 18: This should be `pub(crate)`.

src/util/borrow.rs
```diff
@@ -14,6 +14,15 @@ pub(crate) enum DumbCow<'a, T> {
...
+        match self {
+            DumbCow::Owned(this) => DumbCow::Borrowed(this),
+            DumbCow::Borrowed(this) => DumbCow::Borrowed(this),
+        }
```
BurntSushi, line 22:
```suggestion
        match *self {
            DumbCow::Owned(ref this) => DumbCow::Borrowed(this),
            DumbCow::Borrowed(ref this) => DumbCow::Borrowed(this),
        }
```

src/util/borrow.rs
BurntSushi:
```suggestion
    pub fn borrowed(&self) -> DumbCow<'_, T> {
```

BurntSushi (changes requested):
Nice! Just a few nits.

src/util/borrow.rs
Patryk27: kk, changed

src/util/borrow.rs
Patryk27: kk, changed

BurntSushi (approved):
Thanks!

PR #318, opened 2025-04-11
cargo: restore sanity to iterative development

BurntSushi:
I'm not sure when or how exactly it happened, but in the last weeks,
I've noticed that `rustc` gets effectively stun-locked whenever I make a
change to a source file in Jiff. A quick examination of what the fuck my
computer is doing seems to reveal that it's spending oodles of time
compiling diesel over and over.

I have no idea why this is happening and I don't really care to spend
the time unraveling the mysteries of diesel.

So I took a hammer to the problem. I have effectively shunted all
examples and all "integration" crates out of Jiff's core workspace and
into their own little bloated fiefdoms. To compensate for the fact that
`cargo test --all` no longer tests these things, I've added shell
scripts to run the requisite tests. And those shell scripts are now run
in CI.

I'm now back to a state where I can save a file in Jiff and I get
sub-second `cargo check` response times.

### DIFF
--- a/scripts/test-integrations
+++ b/scripts/test-integrations
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# This script tests "integration" crates explicitly.
+#
+# This isn't done as part of, e.g., `cargo build --all`, because we split these
+# specific crates out of the main workspace to keep dependency trees smaller.
+
+set -e
+
+# cd to the directory containing this crate's Cargo.toml so that we don't need
+# to pass --manifest-path to every `cargo` command.
+cd "$(dirname "$0")/.."
+
+integrations=(
+  "jiff-diesel mysql"
+  "jiff-diesel postgres"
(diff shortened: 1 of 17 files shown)


scripts/test-integrations
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
```
tristan957, line 1: Drive-by review and really nothing big. The shebang that should be used for bash is `#!/usr/bin/env bash`.

BurntSushi: Oh I'm very aware of that! But portability isn't really my concern here to be honest. And if tooling like shellcheck doesn't complain about this, then it's in practice not really something I'll persistently fix. Now if someone tries running these scripts and they fail for this reason, I'm happy to accept a patch.

tristan957: Makes sense!

PR #319, opened 2025-04-11
Use UTC timezone then TZ env var is empty

### LINKED ISSUE #311: an empty but set `TZ` environment variable should probably behave as if `TZ=UTC`
Ref https://github.com/chronotope/chrono/issues/1690
Ref https://github.com/uutils/coreutils/issues/7498

Apparently this is how GNU date works too:

```
$ TZ= date
Sat Apr  5 01:07:44 PM UTC 2025
```

`date` on macOS does the same thing:

```
$ TZ= date
Sat Apr  5 13:12:42 UTC 2025
```

I kinda think that an empty environment variable and one that isn't set should probably behave the same, but this is far from a consistent convention in the Unix world. And it probably makes sense to unify with existing practice (since ultimately that's why Jiff supports `TZ` in the first place).

joeydewaal:
This PR makes an empty `TZ` environment variable act like `TZ=UTC`. Should there be a unit test for this?

Fixes #311

### DIFF
--- a/src/tz/system/mod.rs
+++ b/src/tz/system/mod.rs
@@ -175,7 +175,9 @@ fn get_env_tz(db: &TimeZoneDatabase) -> Result<Option<TimeZone>, Error> {
 
     let Some(tzenv) = std::env::var_os("TZ") else { return Ok(None) };
     if tzenv.is_empty() {
-        // It is commonly agreed (but not standard) that setting an empty `TZ=` uses UTC.
+        // It is commonly agreed (across GNU and BSD tooling at least),
+        // but not standard, that setting an empty `TZ=` is indistinguishable
+        // from `TZ=UTC`.
         return Ok(Some(TimeZone::UTC));
     }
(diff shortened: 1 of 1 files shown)


src/tz/system/mod.rs
```diff
@@ -175,7 +175,8 @@ fn get_env_tz(db: &TimeZoneDatabase) -> Result<Option<TimeZone>, Error> {
 
     let Some(tzenv) = std::env::var_os("TZ") else { return Ok(None) };
     if tzenv.is_empty() {
-        return Ok(None);
+        // It is commonly agreed (but not standard) that setting an empty `TZ=` uses UTC.
```
BurntSushi, line 178:
```suggestion
        // It is commonly agreed (across GNU and BSD tooling at least),
        // but not standard, that setting an empty `TZ=` is indistinguishable
        // from `TZ=UTC`.
```

Slight re-phrase and wrap to 79 columns (inclusive).

PR #340, opened 2025-04-26
fmt/rfc2822: fix handling whitespace at comma and time separators

### LINKED ISSUE #339: support more corner cases in RFC 2822 parsing
> @BurntSushi I did find some test cases that seem valid to me, but that Jiff doesn't parse. They all concern whitespace patterns technically allowed, but of course unlikely in practice:
> 
> - `Sat , 15 Aug 2020 23:12:09 MST` (whitespace before comma is allowed since `obs-day-of-week = [CFWS] day-name [CFWS]
> `)
> - `Sat, 15 Aug 2020 23 :12 : 09 MST` (whitespace before/after each time component is allowed, e.g. `obs-minute = [CFWS] 2DIGIT [CFWS]`)
> - `Sat,15 Aug 2020 23:12:09 MST`(whitespace after comma is optional since `day  = ([FWS] 1*2DIGIT) / obs-day`
> - `Sat   ,15 Aug 2020 23:12:09 MST` (combination of the above)
>  

 _Originally posted by @ariebovenberg in [#200](https://github.com/ariebovenberg/whenever/issues/200#issuecomment-2832123414)_

ariebovenberg:
RFC2822 allows some unusual whitespace configurations which weren't covered previously. In particular, these two cases are now handled properly according to spec:

(1) Whitespace is allowed, but may also be completely omitted, on both sides of the `,` following the weekday;
(2) Whitespace is allowed around the time separators (`:`).

Fixes #339

---

# Implementation notes

1. I changed `skip_whitespace` to return a `bool` whether whitespace was actually encountered. This saved some extra checks in parsing the time, and made `parse_whitespace` slightly cleaner.
1. I'm not that familiar with advanced error handling in Rust (`.with_context`), but I tried my best to keep the same errors as before.
1. I added an extra test to ensure there's whitespace between the time (in minute precision) and timezone. This is easy to miss in the implementation now that whitespace after minute is sometimes required, sometimes not. This test wasn't a good fit for any test method, so I put it in `err_parse_invalid`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ Add a "lenient" mode for `strtime` formatting APIs that ignores most errors.
 
 Bug fixes:
 
+* [#340](https://github.com/BurntSushi/jiff/issues/340):
+Allow whitespace in more places in RFC 2822 parser (improves spec compliance).
 * [#346](https://github.com/BurntSushi/jiff/issues/346):
 `TimeZone::get("UTC")` should now always return `TimeZone::UTC`.
 
--- a/src/fmt/rfc2822.rs
+++ b/src/fmt/rfc2822.rs
@@ -442,18 +442,32 @@ impl DateTimeParser {
         let Parsed { value: year, input } = self.parse_year(input)?;
 
         let Parsed { value: hour, input } = self.parse_hour(input)?;
+        let Parsed { input, .. } = self.skip_whitespace(input);
         let Parsed { input, .. } = self.parse_time_separator(input)?;
+        let Parsed { input, .. } = self.skip_whitespace(input);
         let Parsed { value: minute, input } = self.parse_minute(input)?;
+
+        let Parsed { value: whitespace_after_minute, input } =
+            self.skip_whitespace(input);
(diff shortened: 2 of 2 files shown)


BurntSushi (approved):
This looks perfect. Really well done! Thank you!!!

BurntSushi:
I pushed some minor style tweaks and a CHANGELOG update.

> I'm not that familiar with advanced error handling in Rust (.with_context), but I tried my best to keep the same errors as before.

It's not too clever. The `with_context(...)` pattern is [borrowed from `anyhow`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html). Jiff doesn't use `anyhow` directly, but its `Error` type is similar in spirit (with a dumber implementation). The idea is that it just allows chaining errors. It makes composition nearly frictionless.

Some might say that Jiff relies way too heavily on stringly-typed errors, and I would perhaps agree with them. The main alternative is to use structured error types everywhere. I settled on strings instead because I felt that it empowered me to make _good_ error messages with frictionless composition. Where as with structured errors (which is how I initially started), I found myself constantly fighting composition and sacrificing error message quality just to make the code work.

PR #360, opened 2025-05-03
fmt/rfc2822: fix panic in handling of whitespace

### LINKED ISSUE #359: RFC 2822 parser can panic on some inputs
The panic was introduced in #340.

Specifically, the parser will panic on `Wed `. The parser assumes there is a byte after it has skipped whitespace.

BurntSushi:
Skipping whitespace happens immediately before `input[0]`. But skipping
the whitespace could lead to `input` being empty.

We handle the empty case and add a regression test.

Fixes #359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,18 @@
 # CHANGELOG
 
-0.2.12 (TBD)
-============
-TODO
+0.2.12 (2025-05-03)
+===================
+This release fixes a couple bugs in Jiff's parser. In particular, a regression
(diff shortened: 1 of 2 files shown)


ariebovenberg:
Well dang...looks like my latest changes introduced this bug in the parser 😮‍💨 . Really shows the power of fuzz testing though 👀

BurntSushi:
Yeah it was a tricky one. And I missed it in review. I hope to setup fuzzing for Jiff soonish.

PR #370, opened 2025-05-09
tz: improve performance of initial `Zoned::now()`

### LINKED ISSUE #366: Very slow loading of zone info
On GitHub's Ubuntu systems, the following code:

```rust
let _ = jiff::Zoned::now();
```

is taking up to 15 seconds to run.

This is with jiff 0.2.13. Importantly the default features need to be enabled. I tried various permutations of features, but couldn't quite pin down which combination caused the worse behavior. For example:

* tz-system+tzdb-zoneinfo: 1 to 2 seconds
* tz-system+tz-fat+tzdb-bundle-platform: 160µs
* default: 1 to 15 seconds

This is running on `ubuntu-24.04-arm` or `ubuntu-24.04`. The arm system seems much worse, but I suspect that is because the arm machines are just naturally slower.

I don't know if this is expected. If it is, perhaps there could be some warnings in the docs somewhere (apologies if I missed it or this is a dupe issue).

BurntSushi:
Specifically, this improves performance in the case of
`/usr/share/zoneinfo` being on a slow file system. Namely, #366 reports
cases (in CI) where this initialization could take multiple seconds.

In this commit, we remove the file reads---but keep the directory
traversal---for validating files in `/usr/share/zoneinfo` as TZif data.
Namely, it is common that while _most_ files in a zoneinfo directory are
valid TZif, not all are.

The reason for keeping the initial directory traversal is that it
simplifies the implementation for case insensitive time zone identifier
lookups. Without the IDs in memory, doing a case insensitive lookup on
the file system becomes much harder. Especially since IANA time zone
identifiers use mixed case and I believe there is no algorithmic way to
normalize them. If we do need to get rid of the initial directory scan,
we'll likely need to reframe the `TimeZoneDatabase::available()` API as
potentially returning strings that aren't valid identifiers, or do some
kind of data dependent normalization techniques. (Which may need to be
updated as new releases of tzdb come out.)

In any case, removing the 4-byte file reads seems to dramatically cut
down initialization time. From multiple seconds to 33ms in the slowest
case. In particular, the directory traversal is now done in a way where
the number of syscalls on Unix should be about equivalent to the number
of directories in `/usr/share/zoneinfo` and *not* the number of files.

We also add a test for a slow initial `Zoned::now()` call.

Fixes #366

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -367,6 +367,28 @@ jobs:
     - name: Build all examples
       run: ./scripts/test-integrations
 
+  # A test to ensure that `Zoned::now()` doesn't take "too long" on its
+  # first run.
+  time-zone-init:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+        - os: ubuntu-latest
+        - os: ubuntu-24.04-arm
+        - os: macos-latest
+        - os: windows-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: stable
+    - run: cargo test --test integration init::zoned_now -- --nocapture
+
+
   # Generic testing for most cross targets. Some get special treatment in
   # other jobs.
   cross:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Bug fixes:
 * [#365](https://github.com/BurntSushi/jiff/issues/365):
 Fixes a compile error in Jiff when only the `tzdb-concatenated` feature was
 enabled.
+* [#366](https://github.com/BurntSushi/jiff/issues/366):
+Fixes slow initial `Zoned::now()` in environments where `/usr/share/zoneinfo`
+is on a very slow file system (like CI environments).
 
 
 0.2.13 (2025-05-05)
--- a/src/tz/db/mod.rs
+++ b/src/tz/db/mod.rs
@@ -771,4 +771,63 @@ mod tests {
         assert_eq!(db.get("Etc/Unknown").unwrap(), TimeZone::unknown());
         assert_eq!(db.get("etc/UNKNOWN").unwrap(), TimeZone::unknown());
     }
+
+    /// This checks that our zoneinfo database never returns a time zone
+    /// identifier that isn't presumed to correspond to a real and valid
+    /// TZif file in the tzdb.
+    ///
+    /// This test was added when I optimized the initialized of Jiff's zoneinfo
+    /// database. Originally, it did a directory traversal along with a 4-byte
+    /// read of every file in the directory to check if the file was TZif or
+    /// something else. This turned out to be quite slow on slow file systems.
+    /// I rejiggered it so that the reads of every file were removed. But this
+    /// meant we could have loaded a name from a file that wasn't TZif into
+    /// our in-memory cache.
+    ///
+    /// For doing a single time zone lookup, this isn't a problem, since we
+    /// have to read the TZif data anyway. If it's invalid, then we just
+    /// return `None` and log a warning. No big deal.
+    ///
+    /// But for the `TimeZoneDatabase::available()` API, we were previously
+    /// just returning a list of names under the presumption that every such
+    /// name corresponds to a valid TZif file. This test checks that we don't
+    /// emit junk. (Which was in practice accomplished to moving the 4-byte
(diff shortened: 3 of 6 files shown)


BurntSushi:
# Run 1

`ubuntu-latest`:

```
2025-05-09T16:03:06.1610144+00:00[Etc/UTC]
first-run-elapsed-microseconds:388603
2025-05-09T16:03:06.549613231+00:00[Etc/UTC]
second-run-elapsed-microseconds:7
```

`ubuntu-latest-arm`:

```
queued forever
```

`macos-latest`:

```
 2025-05-09T16:03:07.572226+00:00[UTC]
first-run-elapsed-microseconds:60533
2025-05-09T16:03:07.632768+00:00[UTC]
second-run-elapsed-microseconds:6
```

`windows-latest`:

```
2025-05-09T16:05:05.6965758+00:00[Etc/UTC]
first-run-elapsed-microseconds:1531
2025-05-09T16:05:05.6981074+00:00[Etc/UTC]
second-run-elapsed-microseconds:10
```

BurntSushi:
# Run 2

OK, I guess I need to use `ubuntu-24.04-arm` and not `ubuntu-latest`. Wat.
But this does at least reproduce the long timings reported in #366. In
particular, for x86-64 Linux, we get ~2s and for aarch64 Linux, we get
~6s. For macOS, we get 59ms. (I wonder if the M1 runners have SSDs.)

`ubuntu-latest`:

```
2025-05-09T17:02:09.685431847+00:00[Etc/UTC]
first-run-elapsed-microseconds:1936850
2025-05-09T17:02:11.62228347+00:00[Etc/UTC]
second-run-elapsed-microseconds:7
```

`ubuntu-latest-arm`:

```
2025-05-09T17:02:21.821768226+00:00[Etc/UTC]
first-run-elapsed-microseconds:5842296
2025-05-09T17:02:27.664060677+00:00[Etc/UTC]
second-run-elapsed-microseconds:6
```

`macos-latest`:

```
2025-05-09T17:02:04.034992+00:00[UTC]
first-run-elapsed-microseconds:59888
2025-05-09T17:02:04.094861+00:00[UTC]
second-run-elapsed-microseconds:6
```

`windows-latest`:

```
2025-05-09T17:03:26.7212275+00:00[Etc/UTC]
first-run-elapsed-microseconds:1452
2025-05-09T17:03:26.7226815+00:00[Etc/UTC]
second-run-elapsed-microseconds:10
```

BurntSushi:
# Run 3

In this run, I removed the 4-byte reads of each file in `/usr/share/zoneinfo`. This isn't quite correct yet since I think names of files that aren't TZif will show up in `TimeZoneDatabase::available()`, but this at least gives us an idea of what perf looks like if we remove the file reads (but keep the initial directory traversal).

`ubuntu-latest`:

```
2025-05-09T17:09:17.79520694+00:00[Etc/UTC]
first-run-elapsed-microseconds:5956
2025-05-09T17:09:17.801157706+00:00[Etc/UTC]
second-run-elapsed-microseconds:6
```

`ubuntu-latest-arm`:

```
2025-05-09T17:09:28.888595009+00:00[Etc/UTC]
first-run-elapsed-microseconds:33234
2025-05-09T17:09:28.921823308+00:00[Etc/UTC]
second-run-elapsed-microseconds:6
```

`macos-latest`:

```
2025-05-09T17:10:26.819157+00:00[UTC]
first-run-elapsed-microseconds:3671
2025-05-09T17:10:26.822808+00:00[UTC]
second-run-elapsed-microseconds:5
```

`windows-latest`:

```
 2025-05-09T17:09:56.9830359+00:00[Etc/UTC]
first-run-elapsed-microseconds:1501
2025-05-09T17:09:56.9845435+00:00[Etc/UTC]
second-run-elapsed-microseconds:15
```

BurntSushi:
Just removing the file read here seems to have dramatically improved things. 6ms on x86-64 and 33ms on aarch64 still isn't _great_, but doing better is I think a fair bit more work. Namely, if we don't scan the directory ahead of time, then doing case insensitive lookups suddenly becomes a lot trickier. It's not an intractable problem, but it's definitely an annoying one. Tools like GNU date work around this by declining to support case insensitive lookups. [Temporal requires case insensitive lookups](https://tc39.es/proposal-temporal/#sec-getavailablenamedtimezoneidentifier). I also think that's the right thing to do for reasons of good sense.

I think what's left here is to make sure `TimeZoneDatabase::available` does the right thing. At present, I think the code treats anything returned from the directory traversal as a purportedly valid time zone. But by removing 4-byte TZif validation check, we allow plausible false positives to sneak through. So that 4-byte check will need to be moved to `TimeZoneDatabase::available` I guess.

Issue #298, opened 2025-03-14
Can `jiff` make it easier to (de)serialize `std::time::Duration` itself?


LukeMathWalker:
I've migrated Pavex, the web framework I'm working on, to `jiff` in [this commit](https://github.com/LukeMathWalker/pavex/pull/470/commits/79bc27575b60c51e215f6f6b2b7f8f8f8588ff76).
Things went pretty smoothly, the documentation was excellent all around.

There was one instance where I thought "this could be better": [this hunk](https://github.com/LukeMathWalker/pavex/pull/470/commits/79bc27575b60c51e215f6f6b2b7f8f8f8588ff76#diff-025ed830736f0c68576a3c9f727500c21aea62f4efb951664d06418641a5bd87R38-R55). Graceful shutdown timeouts, time-to-live for sessions—these are all values that must be modelled as _positive_ durations. I have therefore relied on `std::time::Duration` rather than `jiff`'s `SignedDuration`.
Nonetheless, I want to give users the possibility to pass these configuration values using a human-readable format (e.g. `2 minutes`), a job I used to delegate to the `humantime` crate.

I wonder: would it possible for `jiff` to add facilities to parse `std::time::Duration` in the same way it currently parses `SignedDuration`? 
I'm naively assuming that:
- Most of the parsing logic would be identicalm, sign aside
- It would show up as an additional module (e.g. `std_duration`?) under `jiff::fmt::serde`.

BurntSushi:
I added the "help wanted" label here because I feel like the resolution to this issue has a somewhat clear path. So if someone wanted to work on this, I think I could mentor it. If someone wants to do this, feel free to post questions here or ping me on Discord (handle is `burntsushi5`).

I will do this eventually myself though otherwise!

BurntSushi:
#445 adds support for this. I stopped short of adding a wrapper type for `std::time::Duration` which provides `Display` and `FromStr` impls. However, I did add dedicated parsing/printing routines to the `jiff::fmt::temporal` and `jiff::fmt::friendly` APIs. And it adds Serde helpers to `jiff::fmt::serde::unsigned_duration`.

PR #380, opened 2025-05-23
fmt: add serialization and deserialization for `UnsignedDuration` (std::time::Duration)

aalhendi:
This PR begins to address #298.

### src/fmt/friendly/printer.rs
- Introduces `jiff::fmt::friendly::SpanPrinter::print_unsigned_duration` method which accepts as `core::timeDuration` as `UnsignedDuration`, converts to a `SignedDuration` and reuses the existing `print_duration` method.

### src/fmt/serde.rs
- Add module `jiff::fmt::serde::unsigned_duration` to house serde helpers for `core::time::Duration`
  - `jiff::fmt::serde::unsigned_duration::friendly::compact` holds required and optional serialization functions. Similar to `jiff::fmt::serde::duration::friendly::compact`
  - `jiff::fmt::serde::unsigned_duration::deserialize` holds required and optional deserializers.


BurntSushi:
Thanks, but this PR is pretty messed up.

1. The duration parsing/printing is still going through `SignedDuration`, which means there are some values that won't be handled correctly.
2. One explicit printing method was added, but none for parsing.
3. The `jiff::fmt::serde` helpers for this are not quite structured right.

I think I can probably salvage some parts of this. Thank you for writing docs.

BurntSushi:
Also, the `Deserialize` helpers for `std::time::Duration` don't match the `Deserialize` implementation for `jiff::SignedDuration`. That is, `SignedDuration` supports deserializing _either_ an ISO 8601 duration _or_ a friendly duration. The deserialization helpers for `std::time::Duration` should do the same.

BurntSushi:
Sorry if I didn't make this clear, but I'll take over from here. This is actually quite a bit more difficult task than I thought.

aalhendi:
> Sorry if I didn't make this clear, but I'll take over from here. This is actually quite a bit more difficult task than I thought.

Well received. I had initially committed a hacky/naive fix in hopes of confirming if the general approach was correct and forgot about the PR for a few months. Apologies for that on my end. Looking forward to future Jiff releases!

PR #397, opened 2025-07-30
fmt: Add `BrokenDownTime::set_meridiem`

tiffany352:
Based on the discussion here: https://github.com/tiffany352/rink-rs/pull/210

`set_meridiem` seems to have been left out originally because it's not used for formatting. But it is needed for doing custom parsing, so I'm adding it here.


BurntSushi:
> Should the meridiem should be used during formatting? The `fmt_hour*` functions could be updated to call `hour_ranged()` instead of reading the `hour` field directly.

I think this makes sense. I updated the PR to do exactly this. :-)

PR #404, opened 2025-08-24
Fix minor typo.

dpathakj:


dpathakj:
Looks like the CI failures ("ci / cross") are a setup issue unrelated to this change. Lots of output like this:

```
Test executable failed (exit status: 101).

stderr:

thread 'main' panicked at src/civil/date.rs:11:44:
called `Result::unwrap()` on an `Err` value: failed to find time zone `America/New_York` since there is no time zone database configured
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

BurntSushi:
Yeah it's because doc tests recently started working with cross compilation. I've gotta gate those tests or just not run doc tests with cross in CI. I'll probably go the latter route since it's incredibly slow otherwise.

PR #409, opened 2025-09-14
serde: switch serde dependency to serde_core

paolobarbolini:
[serde_core](https://crates.io/crates/serde_core) has just been released. Crates that don't depend on serde derive macros should use it instead of `serde`, as it speeds up compile times by having the build for the crate itself be independent from `serde-derive` (in case it were to be enabled by some other crate). See the serde_core docs for more info.

On my Ryzen 5 5500U, building this repo with `cargo build --release --features serde` went from 10.83s to 8.76s.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,9 +48,9 @@ default = [
   "tzdb-concatenated",
   "perf-inline",
 ]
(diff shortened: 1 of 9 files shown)


BurntSushi:
Thank you!!!

PR #434, opened 2025-10-18
zoned: add arithmetic impls for `Zoned`

### LINKED ISSUE #396: Consider implementing `std::ops` operators for owned `Zoned` types
Currently symbolic mathematical operations involving `Zoned` types are only implemented for `&Zoned` references (in part because `Zoned` is not `Copy`). This could lead one to assume that these operators are not defined/present at all when attempting to perform these operations owned `Zoned` instances (as would be most natural, for example, when dealing with temporaries).

This would be strictly in addition to the current definitions implemented on `&Zoned` references, to improve the usability of the library and make these features more discoverable (as originally discussed in #390).

(I'm not sure if there are any other types that this proposal should extend to, as I understand it all the remaining important types that one would perform math on are `Copy`?)

BurntSushi:
Previously, we only had impls for `&Zoned`. These impls are a little
weird because `Zoned` isn't `Copy`. So using them will result in
consuming the `Zoned` value. But if that's okay, then having these impls
is a nice quality of life improvement instead of a papercut. In the
worst case, Rust programmers will need to add a `&` to avoid consuming
the `Zoned` value. But I think Rust programmers are pretty used to that.

Fixes #396

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -369,7 +369,7 @@ use jiff::civil::date;
 
 let zdt1 = date(2020, 8, 26).at(6, 27, 0, 0).in_tz("America/New_York")?;
 let zdt2 = date(2023, 12, 31).at(18, 30, 0, 0).in_tz("America/New_York")?;
-let span = &zdt2 - &zdt1;
+let span = zdt2 - zdt1;
 assert_eq!(format!("{span:#}"), "29341h 3m");
 
 # Ok::<(), Box<dyn std::error::Error>>(())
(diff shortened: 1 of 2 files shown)


tisonkun:
Although this is understandable, it can affect downstream projects to encounter compile errors like: https://github.com/fast/logforth/actions/runs/19172497177/job/54808534516

It can be worked around like https://github.com/fast/logforth/pull/199, but it may still be worth a notice.

tisonkun:
> Rust programmers will need to add a & to avoid consuming
> the Zoned value. But I think Rust programmers are pretty used to that.

Found it and YES. But it seems to be missed in the CHANGELOG.

BurntSushi:
What is the actual error? No impls were removed here. I'm guessing a type inference regression?

tisonkun:
Yes. You can see the links above. `let end_time = cur_time + span` would now consume `cur_time` by inference.

PR #443, opened 2025-11-06
docs: typo fixing

jessekrubin:
I noticed some typos in the docs for jiff when reading through the docs so I thought it would be pretty straight forward to fix those...

What I did:

1. Installed `typos-cli` and `codespell` (things I have used the past)
2. Ran both tools to create a list of all words those tools didn't like, and removed the words that probably should not be fixed/altered
3. Ran both tools to fix typos ignoring the list of words from the previous step
4. Selectively added changes that seemed good to go
5. Created this PR

FWIW: I am awful at spelling

### DIFF
--- a/crates/jiff-static/src/shared/util/itime.rs
+++ b/crates/jiff-static/src/shared/util/itime.rs
@@ -320,7 +320,7 @@ impl IDate {
 
     /// Returns the date corresponding to the day of the given year. The day
     /// of the year should be a value in `1..=365`, with February 29 being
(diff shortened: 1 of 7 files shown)


BurntSushi (approved):
Thank you! The number of transpositions I make is horrifying.

jessekrubin:
I totalyl gte whta youer sayign. 

I do it all the time too. I also find errant `:wq`s ALL over things I have worked on.

Issue #412, opened 2025-09-17
impl `Display`, `FromStr`, `Serialize` and `Deserialize` for `ISOWeekDate`


janTatesa:
And possibly other struct which don't have serde support

BurntSushi:
I think this is probably reasonable, since the week date format is specified by ISO 8601.

Today, you have to use the `strftime` and `strptime` APIs to deal with it: https://docs.rs/jiff/latest/jiff/fmt/index.html#bespoke-datetime-formats-via-strptime-and-strftime

But I think adding `Display`, `FromStr`, `Serialize` and `Deserialize` impls on `jiff::civil::ISOWeekDate` makes sense.

If there are other types you want serde support for, please open issues for them and provide a use case.

(Also, this is not "derive serde." Jiff very intentionally does not use serde derives.)

jessekrubin:
FWIW I **VERY** crudely did this here: https://github.com/jessekrubin/ry/blob/main/crates/ryo3-jiff/src/isoformat.rs

Very in favor of this being added to jiff!

aznszn:
im working on this, will make a PR soon

PR #444, opened 2025-11-07
added `FromStr` and `Display` impls for ISOWeekDate

aznszn:
Makes progress on #412

### changes
- Added `Display` and `FromStr` impls for `ISOWeekDate`
- Added `iso_week_date` module in fmt, kept the structure similar to `fmt::rfc2822`
### Issues
I copied over `parse_year`, `parse_year_sign`, and `parse_date_separator` from `fmt::temporal::parser` since the logic is identical and I was not sure how to keep it DRY

with these changes implementing `Deserialize` and `Serialize` will be straightforward so I can do that once this PR is greenlit

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # CHANGELOG
 
+0.2.17 (TBD)
+============
+TODO
+
+Enhancements:
+
+* [#412](https://github.com/BurntSushi/jiff/issues/412):
+Add `Display`, `FromStr`, `Serialize` and `Deserialize` trait implementations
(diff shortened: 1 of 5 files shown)


BurntSushi:
Thanks! I think the code duplication here is a bit much for me to stomach. And I'm not sure an entire new module in `jiff::fmt` is really worth it here. I did some surgery here and moved the parser into the `jiff::fmt::temporal` module to make code reuse simpler, even though a week date format isn't specified by the Temporal spec.

I also added the Serde `Serialize` and `Deserialize` impls, along with some docs on `ISOWeekDate` describing its `FromStr` and `Display` impls.

PR #449, opened 2025-11-26
shared: attempt a faster date <-> rate die algorithm

BurntSushi:
This comes from:
https://www.benjoffe.com/fast-date-64

More specifically, the code is here:
https://github.com/benjoffe/fast-date-benchmarks/blob/7fcf82b07d340ddbec866e15cfe333485439ab7f/algorithms/benjoffe_fast64.hpp

But the benchmarks don't show an improvement:

```
$ critcmp base x01
group                                                   base                                   x01
-----                                                   ----                                   ---
civil_datetime/to_timestamp_static/bundled/jiff         1.00     10.5±0.09ns        ? ?/sec    1.03     10.9±0.07ns        ? ?/sec
civil_datetime/to_timestamp_static/zoneinfo/jiff        1.00     10.4±0.09ns        ? ?/sec    1.03     10.8±0.07ns        ? ?/sec
timestamp/to_civil_datetime_offset_conversion/jiff      1.00      4.4±0.05ns        ? ?/sec    1.03      4.6±0.03ns        ? ?/sec
```

I ran the benchmarks like this:

```
cd bench
```

Before the change:

```
cargo bench --'(civil_datetime/to_timestamp_static|timestamp/to_civil_datetime_offset_conversion).*jiff' --save-baseline base
```

And then after the change:

```
cargo bench --'(civil_datetime/to_timestamp_static|timestamp/to_civil_datetime_offset_conversion).*jiff' --save-baseline x01
```

Then I used [`critcmp`] to compare them:

```
critcmp base x01
```

It's very possible I didn't port it correctly. I haven't scrutinized the
codegen. It's also possible that there is an improvement, but that it's
hard to write a benchmark using Jiff APIs to observe it.

(Note that I left out the ARM-specific bits. I'm testing this on x86-64.
I wanted to test there first before digging into the platform specific
optimizations.)

[`critcmp`]: https://github.com/BurntSushi/critcmp

### DIFF
--- a/crates/jiff-static/src/shared/util/itime.rs
+++ b/crates/jiff-static/src/shared/util/itime.rs
@@ -173,31 +173,40 @@ impl IEpochDay {
     #[cfg_attr(feature = "perf-inline", inline(always))]
     #[allow(non_upper_case_globals, non_snake_case)] // to mimic source
     pub(crate) const fn to_date(&self) -> IDate {
-        const s: u32 = 82;
-        const K: u32 = 719468 + 146097 * s;
-        const L: u32 = 400 * s;
+        // Ported from:
+        // https://github.com/benjoffe/fast-date-benchmarks/blob/7fcf82b07d340ddbec866e15cfe333485439ab7f/algorithms/benjoffe_fast64.hpp#L73
 
-        let N_U = self.epoch_day as u32;
-        let N = N_U.wrapping_add(K);
+        const ERAS: u32 = 14704;
+        const D_SHIFT: u32 = 146097 * ERAS - 719469;
+        const Y_SHIFT: u32 = 400 * ERAS - 1;
+        const SCALE: u32 = 32;
+        const SHIFT_0: u32 = 30556 * SCALE;
+        const SHIFT_1: u32 = 5980 * SCALE;
+        const C1: u64 = 505054698555331; // floor(2^64*4/146097):
+        const C2: u64 = 50504432782230121; // ceil(2^64*4/1461):
+        const C3: u64 = 8619973866219416 * 32 / (SCALE as u64); // floor(2^64/2140):
 
-        let N_1 = 4 * N + 3;
-        let C = N_1 / 146097;
-        let N_C = (N_1 % 146097) / 4;
+        let day_number = self.epoch_day as u32;
+        let rev = D_SHIFT.wrapping_sub(day_number) as u64;
+        let cen = ((C1 as u128) * (rev as u128) >> 64) as u64;
+        let jul = rev - cen / 4 + cen;
 
-        let N_2 = 4 * N_C + 3;
-        let P_2 = 2939745 * (N_2 as u64);
-        let Z = (P_2 / 4294967296) as u32;
-        let N_Y = (P_2 % 4294967296) as u32 / 2939745 / 4;
-        let Y = 100 * C + Z;
+        let num = (C2 as u128) * (jul as u128);
+        let yrs = Y_SHIFT.wrapping_sub((num >> 64) as u32);
+        let low = num as u64;
+        let ypt = ((24451 * SCALE as u128) * (low as u128) >> 64) as u32;
 
-        let N_3 = 2141 * N_Y + 197913;
-        let M = N_3 / 65536;
-        let D = (N_3 % 65536) / 2141;
+        let bump = ypt < 3952 * SCALE;
+        let shift = if bump { SHIFT_1 } else { SHIFT_0 };
 
-        let J = N_Y >= 306;
-        let year = Y.wrapping_sub(L).wrapping_add(J as u32) as i16;
-        let month = (if J { M - 12 } else { M }) as i8;
+        let N = (yrs % 4) * (16 * SCALE) + shift - ypt;
+        let M = N / (2048 * SCALE);
+        let D = ((C3 as u128) * ((N % (2048 * SCALE)) as u128) >> 64) as u32;
+
+        let month = M as i8;
         let day = (D + 1) as i8;
+        let year = yrs.wrapping_add(bump as u32) as i16;
+
         IDate { year, month, day }
     }
 
@@ -351,26 +360,23 @@ impl IDate {
     #[cfg_attr(feature = "perf-inline", inline(always))]
     #[allow(non_upper_case_globals, non_snake_case)] // to mimic source
     pub(crate) const fn to_epoch_day(&self) -> IEpochDay {
-        const s: u32 = 82;
-        const K: u32 = 719468 + 146097 * s;
-        const L: u32 = 400 * s;
+        // Ported from:
+        // https://github.com/benjoffe/fast-date-benchmarks/blob/7fcf82b07d340ddbec866e15cfe333485439ab7f/algorithms/benjoffe_fast64.hpp#L130
 
         let year = self.year as u32;
         let month = self.month as u32;
         let day = self.day as u32;
 
-        let J = month <= 2;
-        let Y = year.wrapping_add(L).wrapping_sub(J as u32);
-        let M = if J { month + 12 } else { month };
-        let D = day - 1;
-        let C = Y / 100;
+        let bump = month <= 2;
+        let yrs = year.wrapping_add(5880000).wrapping_sub(bump as u32);
+        let cen = yrs / 100;
+        let shift = if bump { 8829 } else { -2919 };
 
-        let y_star = 1461 * Y / 4 - C + C / 4;
-        let m_star = (979 * M - 2919) / 32;
-        let N = y_star + m_star + D;
+        let year_days = yrs * 365 + yrs / 4 - cen + cen / 4;
+        let month_days = ((979 * (month as i32) + shift) / 32) as u32;
+        let epoch_day =
+            (year_days + month_days + day).wrapping_sub(2148345369) as i32;
 
-        let N_U = N.wrapping_sub(K);
-        let epoch_day = N_U as i32;
         IEpochDay { epoch_day }
     }
 
--- a/src/shared/util/itime.rs
+++ b/src/shared/util/itime.rs
@@ -171,31 +171,40 @@ impl IEpochDay {
     #[cfg_attr(feature = "perf-inline", inline(always))]
     #[allow(non_upper_case_globals, non_snake_case)] // to mimic source
     pub(crate) const fn to_date(&self) -> IDate {
-        const s: u32 = 82;
-        const K: u32 = 719468 + 146097 * s;
-        const L: u32 = 400 * s;
+        // Ported from:
+        // https://github.com/benjoffe/fast-date-benchmarks/blob/7fcf82b07d340ddbec866e15cfe333485439ab7f/algorithms/benjoffe_fast64.hpp#L73
 
-        let N_U = self.epoch_day as u32;
-        let N = N_U.wrapping_add(K);
+        const ERAS: u32 = 14704;
+        const D_SHIFT: u32 = 146097 * ERAS - 719469;
+        const Y_SHIFT: u32 = 400 * ERAS - 1;
+        const SCALE: u32 = 32;
+        const SHIFT_0: u32 = 30556 * SCALE;
+        const SHIFT_1: u32 = 5980 * SCALE;
+        const C1: u64 = 505054698555331; // floor(2^64*4/146097):
+        const C2: u64 = 50504432782230121; // ceil(2^64*4/1461):
+        const C3: u64 = 8619973866219416 * 32 / (SCALE as u64); // floor(2^64/2140):
 
-        let N_1 = 4 * N + 3;
-        let C = N_1 / 146097;
-        let N_C = (N_1 % 146097) / 4;
+        let day_number = self.epoch_day as u32;
+        let rev = D_SHIFT.wrapping_sub(day_number) as u64;
+        let cen = ((C1 as u128) * (rev as u128) >> 64) as u64;
+        let jul = rev - cen / 4 + cen;
 
-        let N_2 = 4 * N_C + 3;
-        let P_2 = 2939745 * (N_2 as u64);
-        let Z = (P_2 / 4294967296) as u32;
-        let N_Y = (P_2 % 4294967296) as u32 / 2939745 / 4;
-        let Y = 100 * C + Z;
+        let num = (C2 as u128) * (jul as u128);
+        let yrs = Y_SHIFT.wrapping_sub((num >> 64) as u32);
+        let low = num as u64;
+        let ypt = ((24451 * SCALE as u128) * (low as u128) >> 64) as u32;
 
-        let N_3 = 2141 * N_Y + 197913;
-        let M = N_3 / 65536;
-        let D = (N_3 % 65536) / 2141;
+        let bump = ypt < 3952 * SCALE;
+        let shift = if bump { SHIFT_1 } else { SHIFT_0 };
 
-        let J = N_Y >= 306;
-        let year = Y.wrapping_sub(L).wrapping_add(J as u32) as i16;
-        let month = (if J { M - 12 } else { M }) as i8;
+        let N = (yrs % 4) * (16 * SCALE) + shift - ypt;
+        let M = N / (2048 * SCALE);
+        let D = ((C3 as u128) * ((N % (2048 * SCALE)) as u128) >> 64) as u32;
+
+        let month = M as i8;
         let day = (D + 1) as i8;
+        let year = yrs.wrapping_add(bump as u32) as i16;
+
         IDate { year, month, day }
     }
 
@@ -349,26 +358,23 @@ impl IDate {
     #[cfg_attr(feature = "perf-inline", inline(always))]
     #[allow(non_upper_case_globals, non_snake_case)] // to mimic source
     pub(crate) const fn to_epoch_day(&self) -> IEpochDay {
-        const s: u32 = 82;
-        const K: u32 = 719468 + 146097 * s;
-        const L: u32 = 400 * s;
+        // Ported from:
+        // https://github.com/benjoffe/fast-date-benchmarks/blob/7fcf82b07d340ddbec866e15cfe333485439ab7f/algorithms/benjoffe_fast64.hpp#L130
 
         let year = self.year as u32;
         let month = self.month as u32;
         let day = self.day as u32;
 
-        let J = month <= 2;
-        let Y = year.wrapping_add(L).wrapping_sub(J as u32);
-        let M = if J { month + 12 } else { month };
-        let D = day - 1;
-        let C = Y / 100;
+        let bump = month <= 2;
+        let yrs = year.wrapping_add(5880000).wrapping_sub(bump as u32);
+        let cen = yrs / 100;
+        let shift = if bump { 8829 } else { -2919 };
 
-        let y_star = 1461 * Y / 4 - C + C / 4;
-        let m_star = (979 * M - 2919) / 32;
-        let N = y_star + m_star + D;
+        let year_days = yrs * 365 + yrs / 4 - cen + cen / 4;
+        let month_days = ((979 * (month as i32) + shift) / 32) as u32;
+        let epoch_day =
+            (year_days + month_days + day).wrapping_sub(2148345369) as i32;
 
-        let N_U = N.wrapping_sub(K);
-        let epoch_day = N_U as i32;
         IEpochDay { epoch_day }
     }
 


BurntSushi:
cc @benjoffe - I hope you don't mind the ping. :-)

benjoffe:
Hi, thanks for trying out the algorithm!

I'm very keen to help out, it'll be great to see this used successfully in the wild.

I have to first note, I have never programmed in Rust before, however it's been high on my list to try out, and this is the perfect push for me.

At first glance, it looks right, and if it passes tests I assume it's been ported correctly.
1. perhaps rust is not pre-compiling constants such as `2048 * SCALE`?
2. perhaps the 128-bit arithmetic does not compile directly to raw 64-bit register access without extra work?
3. perhaps the line `let shift = if bump ... `is compiling to an actual branch?

And finally, perhaps you are testing this on a x64 Mac? There is a note at the end of my blog post that **x64** Mac does not surface the same speed gains in the benchmarks (it's still faster, but by less), which I presume is due to the power management features of the OS or chip activating when 64 bit math is run in a tight loop.

If it turns out that Rust is unable to compile the 128-bit arithmetic to direct register access, then it might be that the 32-bit friendly alternative is going to end up faster in this language*.  It would be unfortunate to have to fallback to that, as it only achieves 3/4 of the speed gains over Neri-Schneider, but could be an interesting thing to test if you run out of ideas.

* https://github.com/benjoffe/fast-date-benchmarks/blob/main/algorithms/benjoffe_fast32.hpp

Finally, I note that the inverse function can be 1-cycle faster if you only need to support 32-bit inputs but are only targeting speed on 64-bit machines - by using 64-bit internal arithmetic and then changing `yrs * 365 + yrs / 4` to `yrs * 1461 / 4`.

I'll be able to have a deeper dive later, hopefully in the next week or two.

BurntSushi:
I don't have hands on a keyboard, so excuse the short reply. But thank you for responding!

rustc uses llvm, so I would be very surprised if rudimentary constant folding or simple branch avoidance wasn't happening. The 128-bit handling I'm less sure about though. For your benchmarks, did you use gcc or clang? Either way, when I get hands on a keyboard, I'll post the codegen here (and show you how I got it). The previous Neri-Schneider implementation used similar techniques, and I had looked at its codegen and at least verified there was no branching or div instructions.

My benchmarks above were on a quiet Linux desktop with the CPU governor set to performance mode.

Thanks again for taking a look!

BurntSushi:
Also, Jiff only requires 32 bit integers to represent rata die. Namely, Jiff only supports years -9999 to 9999.

benjoffe:
The benchmark table at the bottom of the blog post specifies the compilers used. For Windows it was MSVC 19.44, and on Apple M4 Pro: Apple clang 17.0.0.

Since the date range is restricted to just +-9999 - you don't need to hoist to 64-bit to safely use the more compact, and faster code in the inverse: `yrs * 1461 / 4`

benjoffe:
Upon further investigation, perhaps the issue is the usage of (A as u128) * (B as u128).
In the worst case, the compiler might possibly turn this into four u64 bit multiplications and stitch the results together.
There seems to be a feature: core::arch::x86_64::_mulx_u64 which might do this more directly.

benjoffe:
These are the results on my Macbook Pro M4 Pro, following your steps with no code changes:

``` group                                                 base                                   x01
-----                                                 ----                                   ---
civil_datetime/to_timestamp_static/bundled/jiff       1.00      7.7±0.22ns        ? ?/sec    1.03      7.9±0.09ns        ? ?/sec
civil_datetime/to_timestamp_static/zoneinfo/jiff      1.00      7.9±0.09ns        ? ?/sec    1.03      8.1±0.14ns        ? ?/sec
timestamp/to_civil_datetime_offset_conversion/jiff    1.12      3.9±0.06ns        ? ?/sec    1.00      3.5±0.03ns        ? ?/sec```

**Comment by BurntSushi** [comment]
Hmmm, if I do this:

```
diff --git a/src/shared/util/itime.rs b/src/shared/util/itime.rs
index 58efd02..56f3840 100644
--- a/src/shared/util/itime.rs
+++ b/src/shared/util/itime.rs
@@ -370,7 +370,7 @@ impl IDate {
         let cen = yrs / 100;
         let shift = if bump { 8829 } else { -2919 };

-        let year_days = yrs * 365 + yrs / 4 - cen + cen / 4;
+        let year_days = yrs * 1461 / 4 - cen + cen / 4;
         let month_days = ((979 * (month as i32) + shift) / 32) as u32;
         let epoch_day =
             (year_days + month_days + day).wrapping_sub(2148345369) as i32;
```

Then that `yrs * 1461` multiplication overflows.

**Comment by BurntSushi** [comment]
To get the codegen for the YMD -> rate die conversion, you'll need `cargo asm`, which you can get with `cargo install cargo-show-asm`.

First you'll need to mark the function as unlineable:

```
diff --git a/src/shared/util/itime.rs b/src/shared/util/itime.rs
index 58efd02..7c7bf7d 100644
--- a/src/shared/util/itime.rs
+++ b/src/shared/util/itime.rs
@@ -355,7 +355,7 @@ impl IDate {
     /// This is Neri-Schneider. There's no branching or divisions.
     ///
     /// Ref: https://github.com/cassioneri/eaf/blob/684d3cc32d14eee371d0abe4f683d6d6a49ed5c1/algorithms/neri_schneider.hpp#L83
-    #[cfg_attr(feature = "perf-inline", inline(always))]
+    #[inline(never)]
     #[allow(non_upper_case_globals, non_snake_case)] // to mimic source
     pub(crate) const fn to_epoch_day(&self) -> IEpochDay {
         // Ported from:
```

And then run:

```
cargo asm --profile release -p jiff --lib 'shared::util::itime::IDate::to_epoch_day'
```

Which will print the assembly to stdout:

```asm
.section .text.jiff::shared::util::itime::IDate::to_epoch_day,"ax",@progbits
	.globl	jiff::shared::util::itime::IDate::to_epoch_day
	.p2align	4
.type	jiff::shared::util::itime::IDate::to_epoch_day,@function
jiff::shared::util::itime::IDate::to_epoch_day:
	.cfi_startproc
	movsx eax, word ptr [rdi]
	movsx ecx, byte ptr [rdi + 2]
	cmp ecx, 3
	sbb eax, 0
	imul edx, ecx, 979
	lea esi, [rdx - 2919]
	add edx, 8829
	cmp ecx, 3
	cmovae edx, esi
	add eax, 5880000
	imul rcx, rax, 1374389535
	mov rsi, rcx
	shr rsi, 37
	imul r8d, eax, 365
	shr eax, 2
	shr rcx, 39
	lea r9d, [rdx + 31]
	test edx, edx
	cmovns r9d, edx
	movsx edx, byte ptr [rdi + 3]
	sar r9d, 5
	add eax, edx
	add eax, r8d
	sub eax, esi
	add eax, ecx
	add eax, r9d
	add eax, 2146621927
	ret
```

You can also ask for llvm IR:

```
cargo asm --profile release -p jiff --lib --llvm 'shared::util::itime::IDate::to_epoch_day'
```

Which gives:

```
; jiff::shared::util::itime::IDate::to_epoch_day
; Function Attrs: mustprogress nofree noinline norecurse nosync nounwind nonlazybind willreturn memory(argmem: read) uwtable
define noundef range(i32 -12692891, 11253375) i32 @jiff::shared::util::itime::IDate::to_epoch_day(ptr noalias noundef readonly align 2 captures(none) dereferenceable(4) %self) unnamed_addr #22 {
start:
  %_3 = load i16, ptr %self, align 2, !noundef !10
  %0 = getelementptr inbounds nuw i8, ptr %self, i64 2
  %_5 = load i8, ptr %0, align 2, !noundef !10
  %1 = getelementptr inbounds nuw i8, ptr %self, i64 3
  %_7 = load i8, ptr %1, align 1, !noundef !10
  %bump = icmp ult i8 %_5, 3
  %. = select i1 %bump, i32 8829, i32 -2919
  %year = sext i16 %_3 to i32
  %_10 = add nsw i32 %year, 5880000
  %_11.neg = sext i1 %bump to i32
  %yrs = add nsw i32 %_10, %_11.neg
  %cen = udiv i32 %yrs, 100
  %month = sext i8 %_5 to i32
  %day = sext i8 %_7 to i32
  %_17 = mul nuw i32 %yrs, 365
  %_181 = lshr i32 %yrs, 2
  %_19 = udiv i32 %yrs, 400
  %_23 = mul nsw i32 %month, 979
  %_22 = add nsw i32 %_23, %.
  %_21 = sdiv i32 %_22, 32
  %_16 = add nsw i32 %day, 2146621927
  %_15 = add nuw i32 %_16, %_181
  %year_days = add i32 %_15, %_17
  %_28 = sub nsw i32 %year_days, %cen
  %_27 = add nsw i32 %_28, %_19
  %_26 = add nsw i32 %_27, %_21
  ret i32 %_26
}
```

And now using the above approach for the rata die -> YMD conversion (don't forget to mark `IEpochDay::to_date` as unlineable), I use this command:

```
cargo asm --profile release -p jiff --lib 'shared::util::itime::IEpochDay::to_date'
```

Which gives the Assembly:

```
.section .text.jiff::shared::util::itime::IEpochDay::to_date,"ax",@progbits
	.globl	jiff::shared::util::itime::IEpochDay::to_date
	.p2align	4
.type	jiff::shared::util::itime::IEpochDay::to_date,@function
jiff::shared::util::itime::IEpochDay::to_date:
	.cfi_startproc
	mov ecx, -2147476477
	sub ecx, dword ptr [rdi]
	movabs rdx, 505054698555331
	mov rax, rcx
	mul rdx
	add rcx, rdx
	shr edx, 2
	sub rcx, rdx
	movabs rdx, 50504432782230121
	mov rax, rcx
	mul rdx
	mov rcx, rdx
	mov edx, 782432
	mul rdx
	mov esi, 5881599
	sub esi, ecx
	mov ecx, esi
	and ecx, 3
	shl ecx, 9
	sub ecx, edx
	lea eax, [rcx + 977792]
	add ecx, 191360
	cmp edx, 126464
	cmovae ecx, eax
	movzx eax, cx
	adc esi, 0
	movabs rdx, 8619973866219416
	mul rdx
	shl edx, 24
	and ecx, 16711680
	or ecx, edx
	movzx eax, si
	add eax, ecx
	add eax, 16777216
	ret
```

And adding the `--llvm` flag gives this IR:

```
; jiff::shared::util::itime::IEpochDay::to_date
; Function Attrs: mustprogress nofree noinline norecurse nosync nounwind nonlazybind willreturn memory(argmem: read) uwtable
define range(i32 16777216, 536870912) i32 @jiff::shared::util::itime::IEpochDay::to_date(ptr noalias noundef readonly align 4 captures(none) dereferenceable(4) %self) unnamed_addr #22 {
start:
  %_3 = load i32, ptr %self, align 4, !noundef !10
  %_5 = sub i32 -2147476477, %_3
  %rev = zext i32 %_5 to i64
  %_9 = zext i32 %_5 to i128
  %_8 = mul nuw nsw i128 %_9, 505054698555331
  %_7 = lshr i128 %_8, 64
  %cen = trunc nuw nsw i128 %_7 to i64
  %_121 = lshr i64 %cen, 2
  %_11 = add nuw nsw i64 %cen, %rev
  %jul = sub nsw i64 %_11, %_121
  %_14 = zext i64 %jul to i128
  %num = mul nuw nsw i128 %_14, 50504432782230121
  %_22 = and i128 %num, 18446744073709551615
  %_21 = mul nuw nsw i128 %_22, 782432
  %_20 = lshr i128 %_21, 64
  %ypt = trunc nuw nsw i128 %_20 to i32
  %bump = icmp samesign ult i32 %ypt, 126464
  %. = select i1 %bump, i32 191360, i32 977792
  %_17 = lshr i128 %num, 64
  %_16 = trunc i128 %_17 to i32
  %yrs = sub i32 5881599, %_16
  %_28 = shl i32 %yrs, 9
  %_27 = and i32 %_28, 1536
  %_26 = sub nsw i32 %_27, %ypt
  %N = add nsw i32 %_26, %.
  %_34 = and i32 %N, 65535
  %_33 = zext nneg i32 %_34 to i128
  %_32 = mul nuw nsw i128 %_33, 8619973866219416
  %_40 = zext i1 %bump to i32
  %_39 = add i32 %yrs, %_40
  %sh.diff = lshr i128 %_32, 40
  %tr.sh.diff = trunc nuw nsw i128 %sh.diff to i32
  %_37 = and i32 %tr.sh.diff, 520093696
  %_0.sroa.3.0.insert.shift = add nuw nsw i32 %_37, 16777216
  %_0.sroa.2.0.insert.ext = and i32 %N, 16711680
  %_0.sroa.2.0.insert.insert = or disjoint i32 %_0.sroa.3.0.insert.shift, %_0.sroa.2.0.insert.ext
  %_0.sroa.0.0.insert.ext = and i32 %_39, 65535
  %_0.sroa.0.0.insert.insert = or disjoint i32 %_0.sroa.2.0.insert.insert, %_0.sroa.0.0.insert.ext
  ret i32 %_0.sroa.0.0.insert.insert
}
```

BurntSushi:
If I set the environment variable `RUSTFLAGS="-C target-cpu=x86-64-v3"`, then that seems to cause the rata die -> YMD conversion to use `mulx` instructions (which I _think_ is what you're hinting at with the `_mulx_u64` link):

```
.section .text.jiff::shared::util::itime::IEpochDay::to_date,"ax",@progbits
	.globl	jiff::shared::util::itime::IEpochDay::to_date
	.p2align	4
.type	jiff::shared::util::itime::IEpochDay::to_date,@function
jiff::shared::util::itime::IEpochDay::to_date:
	.cfi_startproc
	mov edx, -2147476477
	sub edx, dword ptr [rdi]
	movabs rax, 505054698555331
	mulx rax, rax, rax
	mov ecx, eax
	shr ecx, 2
	add rax, rdx
	sub rax, rcx
	movabs rcx, 50504432782230121
	mov rdx, rax
	mulx rcx, rdx, rcx
	mov eax, 782432
	mulx rdx, rdx, rax
	mov eax, 5881599
	sub eax, ecx
	mov ecx, eax
	and ecx, 3
	shl ecx, 9
	sub ecx, edx
	lea esi, [rcx + 977792]
	add ecx, 191360
	cmp edx, 126464
	cmovae ecx, esi
	movzx edx, cx
	movabs rsi, 8619973866219416
	mulx rdx, rdx, rsi
	adc eax, 0
	shl edx, 24
	and ecx, 16711680
	or ecx, edx
	movzx eax, ax
	add eax, ecx
	add eax, 16777216
	ret
```

Benchmarking with `x86-64-v3` can be done with (being sure to mark these functions as `inline(always)` again):

```
RUSTFLAGS="-C target-cpu=x86-64-v3" cargo bench -- '(civil_datetime/to_timestamp_static|timestamp/to_civil_datetime_offset_conversion).*jiff' --save-baseline x02
```

And I get:

```
$ critcmp base x01 x02
group                                                   base                                   x01                                    x02
-----                                                   ----                                   ---                                    ---
civil_datetime/to_timestamp_static/bundled/jiff         1.00     10.5±0.09ns        ? ?/sec    1.03     10.9±0.07ns        ? ?/sec    1.03     10.8±0.14ns        ? ?/sec
civil_datetime/to_timestamp_static/zoneinfo/jiff        1.00     10.4±0.09ns        ? ?/sec    1.03     10.8±0.07ns        ? ?/sec    1.03     10.7±0.13ns        ? ?/sec
timestamp/to_civil_datetime_offset_conversion/jiff      1.00      4.4±0.05ns        ? ?/sec    1.03      4.6±0.03ns        ? ?/sec    1.01      4.5±0.03ns        ? ?/sec
```

So hmmm, no dice.

I'll pause here for now. I haven't tried the 32-bit algorithms yet.

How does the codegen I've shared look to you? I figure you might be able to spot issues there more easily than I can.

benjoffe:
Thanks for showing how to view the assembly, that helps a lot.

It seems the 64x64 -> 128 multiplications are working fine - but there is a left-shift by 24 which is out of nowhere, and indicates that it might be adjusting the math somewhere to no longer be aligned perfectly into the 64-bit registers.

I'll dig deeper in the next week or two and see what can be changed to keep it closer to the intended output.

benjoffe:
Hi,

Thank you for your patience while I've taken longer than expected to investigate this issue.

Given it's been a while I thought I'd just bump this thread with a small update so you know I haven't moved on.

My previous point about the left-shift by 24 appears to be something of a red-herring, as that is just part of the bit-packing for the struct.

At this point, I'm fairly confident that the cause for the slowdown is due to _register-spill_. My testing indicates that making the new function _NOT_ inline actually speeds it up; but it's still slower than the previous inlined code (as non inlined code involves writing to memory etc.).

If neither the old nor new algorithms are inlined, then the new code seems to clearly win on speed. Given that optimisation settings use inlining everywhere, this may not be a good solution.

Now, the reason the new algorithm is more subject to register spill is because of all the usage of 64-bit multiplications everywhere, they tend to fill up the available registers quickly (each using 2 instead of 1), and the compiler seems to not be smart enough to juggle these registers around efficiently when it's inlined into larger functions.

I have been working on the last few weeks on several 32-bit versions of the new fast date algorithm. These seem to be quite fast still, and have the added benefit of not causing a large slowdown on 32-bit computers (as the 64-bit one does). There will be 3 or 4 variants based on different range requirements, and they seem to eliminate the problem here, while still maintaining more than 50% of the overall speedup.

I'll bump this thread again soon once these algorithms are tested and documented, along with benchmark results in this library.

In the meantime, would you be able to possibly advise me on these two points below?
1. would you expect that inlining is used by most usages of this lib or is it mostly used when benchmarking?
2. does anything else in this codebase have baked-in 64-bit target assumptions or is it fairly agnostic as to target platform? If nothing else uses 64-bit multiplication, perhaps sticking to a 32-bit solution is worthwhile for that reason alone.

Regards,
Ben

BurntSushi:
Thanks for keeping on this!

For (1), users of this crate can disable _aggressive_ inlining. But I'd say that's atypical. The Rust world, like the C++ world, uses inlining quite heavily. That said, we can choose not to inline functions if that makes things faster. Inlining is ultimately in service to the goal of runtime performance (and also possibly code size, in some cases). So if not inlining makes things _overall_ faster, then that's fair game.

For (2), the range of dates in this crate is firmly fixed at `-9999-01-01` to `9999-12-31`. Which means the rata die for the Unix epoch fits into 32 bits. In that sense, no, this crate doesn't use 64-bit for _dates_ specifically. But this crate uses 64-bit (and even 96-bit and 128-bit) in other places.

benjoffe:
Hi,

I'm still very confused about why this is performing slow on Windows. 

The good news is I found a different performance improvement which is in that other pull request I just created now. Although I now see it's failing some kind of CI checks, sorry about that :(

I've had success in having different versions of the RD -> YMD algorithm run the benchmark `timestamp/to_civil_datetime_offset_conversion` 5-15% faster on the following platforms:

- Linux (AMD cpu tested only)
- Mac (Apple Silicon / M4 Pro Chip)
- Mac (2 different Intel x86 chips).

It is on Windows computers where things just don't work as expected, but in those cases I have no problem optimising in C++. I no longer believe it's related to register spill, but to something more broad perhaps.

I have noticed that the `bench/Cargo.toml` file has:

```
[profile.bench]
debug = true

```

I just wanted to double check with you that this is intended, and not likely to distort the results?
Changing it didn't really help, but I'm still very new to rust.

BurntSushi:
`debug = true` is just adding debug symbols. That in turn helps profilers give sensible output. But absolutely feel free to set `debug = false` to test things out.

I don't really have a benchmarking workflow for Windows. So as long as there's not a major regression for Windows, I'm totally okay with that if it means a 5-15% improvement on Unix systems.

BurntSushi:
I'm going to close this out for now because it isn't a priority for me, and I like to keep the PR queue clean (for as long as I can) with things I plan to get merged in some reasonable time window. If I revisit the problem here, I'll post in this PR (or ping you in a new issue or PR). Sorry for dropping the ball here. My focus is on Jiff 1.0 and this isn't a Jiff 1.0 problem. (I may still have some fun on a silly perf whim.)

benjoffe:
Hi, no problem, I thought this was on me to investigate anyway.
I have learned that my prior testing of the fast algorithm fell short. I tested on clang for Apple ARM64 and MSVC on Windows x86-64. 
These show the huge gains, but clang on x86 not so much.
This has lead me to realise how varied compilers and platforms can be: ARM64 MSVC it's 60% faster than before, but Clang x86 it's neck-and-neck. Crazy variance.
I have developed an improved alternative RD->YMD algorithm that I haven't had time to fully document, but if I find it works well in this library, I will make a new pull request.

I have a completely unrelated optimisation that I would like to share soon, on my M4 mac pro it speeds up a certain function by > 30% with just a couple of lines, and hopefully similar elsewhere. I will test on more platforms and document it and share in a new pull request soon.

BurntSushi:
That sounds amazing.

And to be honest, this also seems like a problem with _my_ benchmarking methodology more so than yours. I'm only testing on x86-64. If there isn't an improvement there but there is in other environments (especially aarch64), then that is worth something!

I'll actually re-open this for now since there is a clear and simple actionable item for me: benchmark Jiff in more environments.

I also love your future work idea. :)

benjoffe:
Great!

But definitely there is an issue with the algo, not necessarily your benchmark.

I have learned a lot of tricks since creating that one. The new algo is a bit different (no backward counting), but the same general shape. It seems faster on all platforms I tested, so no point benching the old one any further.

It's very reasonably possible that what's faster in one codebase is not necessarily going to translate the same in another, even with identical assembly output! This seem to be due to the surrounding code, which in best cases can oveelap the execution due to superscalar features.

I can probably share it in the next 24hr in C++ draft form if you are interested to try the port.

Sorry it took a while to share, I've been sidetracked working on other global calendars algorithms, such as Hebrew 4-8x faster than traditional methods, guessing that's quite out of scope here though.

BurntSushi:
> such as Hebrew 4-8x faster than traditional methods, guessing that's quite out of scope here though.

For Jiff yes. But [ICU4X](https://github.com/unicode-org/icu4x) might be very interested in that!

BurntSushi:
> I can probably share it in the next 24hr in C++ draft form if you are interested to try the port.

Yeah! No rush. Not sure when I'll give it a try but doing a little port sounds like fun.

benjoffe:
Hi,

So, I forgot I actually uploaded the improved version to a branch a while ago with a bit of commentary:
https://github.com/benjoffe/fast-date-benchmarks/blob/x64_llvm_improvements/algorithms/benjoffe_fast64_v2.hpp

The 32-bit version is also available here:
https://github.com/benjoffe/fast-date-benchmarks/blob/main/algorithms/benjoffe_fast32_v2.hpp

I'm not convinced it will speed up jiff in your x86 benchmarks, so definitely don't prioritise toying with it unless you're bored! Even if it is faster, it'll probably be marginal overall (other than ARM64/32 for the two versions respectively), there's diminishing returns in optimising this function, I'm a bit crazy to work on it as hard as I did.

Nonetheless, the key changes I find to be quite interesting, and will hopefully update the web page at some point to explain at a future date. Both of these avoid backwards counting by:
* 64-bit version: shifts the domain to the negative date range, and uses signed shift.
* 32-bit version: uses multiplication by negative constants to achieve the same effect.

Essentially, something has to be negative or backwards in order to avoid the "*4+3" terms. These terms really halt ARM pipelines, Apple Silicon in particular seems to be really good at throwing a MUL-SHIFT into a subsequent MUL-SHIFT. Whenever an addition is inserted, it seems to really slow the pipeline.

x86 on the other hand doesn't seem to mind such terms, as they are MUL -> LEA -> MUL, and these chips seem pretty well optimised for that sequence.

The things that help x86 in v2 of the 64-bit version are:
- Shift to negative domain allows moving an addition to the end, where it sits in the shadow of prior ops (this is the source of the minor speedbump for apple silicon as well)
- LEA assembly operation can be used in several places where it could not be used before
- Smaller constants are used throughout to facilitate the LEA. The helps avoid a SCALE factor for platform micro-op, which simplifies the code, and also allows the final 'month' value to be byte-aligned (might speed up libs that return a bit-packed struct, though jiff uses inlining so not applicable I guess).

Finally, it's worth noting that if the calling code of the RD->YMD function can be vectorised, you're probably better of with Neri-Schneider algo, it's more optimal for that I think (not tested though).

Hoping to make progress on the other optimisation pull-request soon.

Cheers,
Ben

BurntSushi:
Thank you!!! And yeah, given Apple Silicon, I basically consider aarch64 performance very close in terms of priority with x86-64 performance. I just don't do enough of it because I don't dogfood aarch64 nearly as much as I should.

PR #469, opened 2026-01-01
timestamp: use `TryFrom` for `SystemTime`

### LINKED ISSUE #467: switch `impl From<Timestamp> for std::time::SystemTime` to `TryFrom`
Specifically, here is the current impl:

https://github.com/BurntSushi/jiff/blob/42d83ff42a42b1cb5ff0be6f56818d4e6c13c5f5/src/timestamp.rs#L2815-L2831

It specifically notes that we assume a `SystemTime` always uses 64 bits. But this is not a documented guarantee. And indeed, it turns out it's not always true: https://github.com/rust-lang/rust/issues/150505

This is a breaking change, so it will need to be part of the Jiff 1.0 release.

iksuddle:
Resolves #467

As mentioned in the issue, we cannot assume that `SystemTime` always uses at least 64 bits.

### DIFF
--- a/src/timestamp.rs
+++ b/src/timestamp.rs
@@ -2813,19 +2813,21 @@ impl<'a> From<&'a Zoned> for Timestamp {
 }
 
 #[cfg(feature = "std")]
-impl From<Timestamp> for std::time::SystemTime {
+impl TryFrom<Timestamp> for std::time::SystemTime {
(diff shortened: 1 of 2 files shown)


BurntSushi:
I'm not ready to merge breaking change patches yet. This has to wait until Jiff 1.0.

PR #479, opened 2026-01-07
feat: implement Display trait for Weekday

aaron-ang:
I'm looking to integrate more of `jiff` into https://github.com/uutils/coreutils. Currently, some tests rely on `chrono` and extracting the weekday string. This feature is not yet available in `jiff` so I decided to add it.

### DIFF
--- a/src/civil/weekday.rs
+++ b/src/civil/weekday.rs
@@ -1,3 +1,5 @@
+use core::fmt;
+
 use crate::{
     error::Error,
     shared::util::itime::IWeekday,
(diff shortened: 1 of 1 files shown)


BurntSushi:
This is intentionally not provided, because Jiff tries not to venture into things that are better in the domain of locales. I'm open to changing that, but I suspect a work around for cases that don't care about locale is simple enough.

The main exception to staying out of locales if Jiff's strftime. From that you can get a weekday string.

aaron-ang:
got it, thanks for the quick response!

Issue #485, opened 2026-01-14
padding with negative values in `strftime` produces an unexpected result


BurntSushi:
Reproduction:

```rust
use jiff::civil;

fn main() -> anyhow::Result<()> {
    let zdt =
        civil::date(-2025, 1, 13).at(5, 0, 0, 0).in_tz("America/New_York")?;
    println!("{}", zdt.strftime("%_6Y"));

    Ok(())
}
```

Has this output:

```
-  2025
```

But it should be:

```
 -2025
```

Note that there is also an issue when zero padding is used. Specifically, this program:

```rust
use jiff::civil;

fn main() -> anyhow::Result<()> {
    let zdt =
        civil::date(-2025, 1, 13).at(5, 0, 0, 0).in_tz("America/New_York")?;
    println!("{}", zdt.strftime("%06Y"));

    Ok(())
}
```

Has this output:

```
-002025
```

But it should be:

```
-02025
```

For reference, GNU date behaves as expected:

```
$ date +'%06Y' -d '@-99999999999'
-01199
$ date +'%_6Y' -d '@-99999999999'
 -1199
```

PR #487, opened 2026-01-14
fix: correct padding with negative values in strftime

ChrisDryden:
This is a proposed fix for https://github.com/BurntSushi/jiff/issues/485 where the padding values when using a space as the padding byte adds the '-' sign before the spaces. This approach tried to separate the logic for adding the padding from the happy path since there has been a bunch of optimization work done to make this faster and its a rare case to have negative numbers. 

The overarching purpose of this fix is to use it in uutils coreutils: https://github.com/uutils/coreutils/issues/10242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ Enable some non-default features for the Rust Playground deployment.
 
 Bug fixes:
 
+* [#485](https://github.com/BurntSushi/jiff/issues/485):
+Fix bug with padding for negative integers in `strftime`.
 * [#486](https://github.com/BurntSushi/jiff/issues/486):
 Make `%^c` result in uppercase strings where appropriate.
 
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thanks! This wasn't quite right as-is, but I think your implementation had the right idea. My guess is that you were biasing toward the existing tests too strongly. The key example is this:

```
$ date +'%Y' -d '@-63999999999'
-059
```

That is, even when a explicit padding isn't set, the `-` is still part of the padding. But in this PR, you were gating the special casing on `self.width.is_some()`, which was only set when there was an explicit padding in the format string. But we want the special handling even when there isn't. This did require changing the result of a few existing tests, but I think that's okay because arguably the result was wrong.

I fixed this up and made a few other minor changes. Thanks again!

BurntSushi:
This PR is on crates.io in `jiff 0.2.20`.

Issue #486, opened 2026-01-14
`%^c` doesn't result in uppercase strings where appropriate


BurntSushi:
This program:

```rust
use jiff::civil;

fn main() -> anyhow::Result<()> {
    let zdt =
        civil::date(-2025, 1, 13).at(5, 0, 0, 0).in_tz("America/New_York")?;
    println!("{}", zdt.strftime("%^c %^a"));

    Ok(())
}
```

Has this output:

```
-2025 M01 13, Mon 05:00:00 MON
```

But it should be:

```
-2025 M01 13, MON 05:00:00 MON
```

The `%^a` demonstrates that Jiff supports the `^` uppercase operator, but it isn't working for `%c`. It's probably getting dropped in the translation somewhere. It works for GNU date:

```
$ date +'%c'
Wed 14 Jan 2026 12:05:32 PM EST
$ date +'%^c'
WED 14 JAN 2026 12:05:35 PM EST
```

For Jiff, `#` also doesn't work (i.e., `%#c` doesn't swap the case just like `%^c` doesn't use uppercase). Interestingly, `%#c` doesn't work for GNU date either:

```
$ date +'%#c'
Wed 14 Jan 2026 12:05:52 PM EST
$ date +'%a'
Wed
$ date +'%#a'
WED
```

I can't tell from `man date` whether this is a bug or not, but it doesn't seem like behavior Jiff should copy.

PR #488, opened 2026-01-14
apply uppercase flag to compound specifiers #486

ChrisDryden:
This approach for https://github.com/BurntSushi/jiff/issues/486 checks the extension flag that is passed into the format_* to change the format string if the uppercase flag was set. This was the main approach that I found could be more easily extended for other compound flags.

The overarching purpose of this fix is to use it in uutils coreutils: https://github.com/uutils/coreutils/issues/10242


BurntSushi:
This PR is on crates.io in `jiff 0.2.20`.

PR #494, opened 2026-02-03
Optimize `Zoned::strftime` to avoid cloning `TimeZone`

Nahor:
Use a reference to `TimeZone` in `BrokenDownTime` instead of a clone.

For performance, adapting the sample code from #492, I get:
```
# before the change
time: 1.6250578s
speed: 30768136 iter/sec

# after the change
time: 194.8673ms
speed: 256584860 iter/sec
```

Unfortunately, this is a breaking change:
```rust
let fmt = { 
    let zdt = date(2026, 1, 1).in_tz("America/Los_Angeles").unwrap();
    zdt.strftime("%Y-%m-%d")
};
let _ = fmt.to_string();
```
will not compile anymore, but I do not expect many applications needing this and it aligns with Rust's `format_args!`, i.e:
```rust
let fmt = {
	let s = std::string::String::from("foobar");
	format_args!("{}", s)
};
let _ = fmt.to_string();
```
does not compile either.

The PR also contains a couple quick fixes for some clippy warnings.

### DIFF
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -170,6 +170,7 @@ use crate::{
 ///
 /// [add-date-rounding]: https://github.com/BurntSushi/jiff/issues/1
 #[derive(Clone, Copy, Hash)]
+#[allow(clippy::derived_hash_with_manual_eq)]
 pub struct Date {
     year: Year,
     month: Month,
--- a/src/civil/iso_week_date.rs
+++ b/src/civil/iso_week_date.rs
@@ -156,6 +156,7 @@ use crate::{
 /// # Ok::<(), Box<dyn std::error::Error>>(())
 /// ```
 #[derive(Clone, Copy, Hash)]
+#[allow(clippy::derived_hash_with_manual_eq)]
 pub struct ISOWeekDate {
     year: ISOYear,
     week: ISOWeek,
--- a/src/tz/timezone.rs
+++ b/src/tz/timezone.rs
@@ -2073,7 +2073,7 @@ mod repr {
                 seconds.checked_shl(4),
                 "offset small enough for left shift by 4 bits",
             );
-            assert!(usize::MAX >= 4_294_967_295);
+            assert!(core::mem::size_of::<usize>() >= 4);
             // usize cast is okay because Jiff requires 32-bit.
             let ptr = without_provenance((shifted as usize) | Repr::FIXED);
             Repr { ptr }


BurntSushi:
This falls into the bucket of "major" breaking changes. We can't add lifetime/type parameters to types in semver compatible releases. And especially not types like `BrokenDownTime`, [which likely appear in user code](https://github.com/BurntSushi/biff/blob/1fd0f0a9b264b0620274c03052006e6132657cdd/src/locale/enabled.rs#L96).

Even putting aside semver compatibility here, I very much do not like parameterizing types like `BrokenDownTime` by a lifetime. It's quite cumbersome and increases API complexity.

I think the better way to fix this is to find a way to make cloning faster in more cases.

Another alternative that I think would be semver compatible is to do the work we want to use the time zone for in the `From` impl. It's only used in one place:

https://github.com/BurntSushi/jiff/blob/3c9ba667390a1f3aeb104fb9110475defa609bb0/src/fmt/strtime/printer.rs#L682-L689

The downside here is that the natural way to go about this at present would be to get the time zone abbreviation and store it as a `Box<str>`. Which is likely cheaper in the benchmark with severe contention, but more costly in the case where there is no contention. And of course, this is also doing a tz lookup, which will penalize any case where `Zoned::strftime` is used without `%Z`.

Nahor:
> We can't add lifetime/type parameters to types in semver compatible releases.

Yes, it is breaking, but better now, when it's still half-expected, than after the 1.0 release, as you mentioned multiple times in the changelogs.
Moreover, can you reasonably guarantee you won't need a lifetime parameter later? The `iana` field for instance feels like a prime-candidate for switching to `&'a str` to support `noalloc`.

> [which likely appear in user code](https://github.com/BurntSushi/biff/blob/1fd0f0a9b264b0620274c03052006e6132657cdd/src/locale/enabled.rs#L96).

A sample of one, from your own code at that, is hardly a fair argument.

There are currently [351 crates](https://crates.io/crates/jiff/reverse_dependencies) depending on `jiff`, which doesn't count non-published projects.
And according to [this](https://sourcegraph.com/search?q=context:global+BrokenDownTime+lang:rust), there are 35-ish mentions of `BrokenDownTime` on GitHub, ignoring `jiff` itself, and `brave-code` which vendored the crate.

I don't know how representative those numbers are, but they don't point to a "likely" usage.

And from your [own words](https://docs.rs/jiff/latest/jiff/fmt/strtime/struct.BrokenDownTime.html):
> This is a lower level aspect of the strptime and strftime APIs that you probably won’t need to use directly

<br/>

> I very much do not like parameterizing types like BrokenDownTime by a lifetime. It's quite cumbersome and increases API complexity.

I understand the reluctance, I do not like having to deal with lifetime either. But lifetimes don't always need to be specified. Only 8 files needed changed and for most, it was only 1 line change. [There](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/mod.rs#L805) [are](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/bench/src/print.rs#L845) [plenty of](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/mod.rs#L466) [other places](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/parse.rs#L1096) in the code where lifetime could be elided, and compiled as they were.

BurntSushi:
> Yes, it is breaking, but better now, when it's still half-expected, than after the 1.0 release, as you mentioned multiple times in the changelogs.

There's probably some miscommunication here? I'm not sure. Nowhere in the CHANGELOG do I suggest merging ["major" breaking changes](https://rust-lang.github.io/rfcs/1105-api-evolution.html#detailed-design) in a semver compatible release.

The CHANGELOG does discuss breaking changes _for the 1.0 release_.

So what I was saying in the first part of my comment was clarifying that this isn't a change that I will be merging for `jiff 0.2.x`. If we were otherwise in alignment on this change, then it would have to wait for the `jiff 1.0` release.

Regardless of anything else, this is not a change I will put in a `jiff 0.2.x` release. There is nothing you can say that will convince me otherwise, so I suggest dropping this thread of discussion.

> Moreover, can you reasonably guarantee you won't need a lifetime parameter later? The `iana` field for instance feels like a prime-candidate for switching to `&'a str` to support `noalloc`.

It certainly won't be _needed_. It would be better to couch this as a discussion of trade-offs. It's possible that the current representation (including the `iana` field) puts a performance ceiling on the API _in exchange_ for API simplicity. Will I regret imposing the ceiling? I'm not sure. It's hard to say. It also depends on how much work a work-around is.

> A sample of one, from your own code at that, is hardly a fair argument.

Fair has nothing to do with it. I regret even bringing it up. Because it doesn't really matter. It's a major breaking change and I will not publish it in a semver compatible release even if I was otherwise on board with it.

> I understand the reluctance, I do not like having to deal with lifetime either. But lifetimes don't always need to be specified. Only 8 files needed changed and for most, it was only 1 line change. [There](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/mod.rs#L805) [are](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/bench/src/print.rs#L845) [plenty of](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/mod.rs#L466) [other places](https://github.com/Nahor/jiff/blob/09ce67856826dd75d7c2b17081240005ac58b26f/src/fmt/strtime/parse.rs#L1096) in the code where lifetime could be elided, and compiled as they were.

I've filed https://github.com/BurntSushi/jiff/issues/495 to track the general thrust of this PR. I do think it's worth doing a careful audit of `BrokenDownTime` more holistically and seeing whether we can do something better here. But for now, I'm going to close this PR because it's a breaking change and I'm not landing breaking changes for Jiff 1.0 right now.

Nahor:
> There's probably some miscommunication here? I'm not sure

Looks like it. I never suggested that this change be pushed in 0.2.x. I never even expected it either because of that API break. But it sounded that it was also a reason to not merge ever, even for a 0.3/0.5/1.0/99.0 version, thus my arguing.

But yeah, this is exhausting. I'm done. There are other cats to skin. GL

PR #501, opened 2026-02-06
doc: update comparison with `time`

### LINKED ISSUE #500: Outdated comparison to `time` crate
- tzdb integration is available via a third-party crate (`time-tz`), though it appears to be unmaintained (I can't speak to this, as it's unofficial)
- The whole "unable to get current local time from multiple threads" is no longer true. I found a workaround that works for any situation where the `TZ` environment variable does not work, which is basically everything. The restriction still applies to _some_ platforms if someone wants to manually refresh the environment, but I don't know of anyone who actually does this (a GitHub search reveals a single instance, though it appears unnecessary).
  I will also note that using Rust's functions for environment variables by itself is **not** sufficient proof of soundness, bizarrely enough. During the initial hubbub when the soundness issue was discovered, I recall a proof of concept that caused a segfault in 100% safe code, relying only on the standard library. I'd have to dig to find this, though.
- `time` has supported `strftime` formats for over a year via a dedicated function. Internally, it uses the same system.
- Performance is almost certainly out of date, as I made significant gains over the past few weeks for parsing and some core algorithms.
  Quickly looking at the benchmarks, I think it would be better of the result were simply returned (and put thru `black_box`, which criterion does with the return value) rather than checking for equality. Otherwise you're benchmarking equality on top of what you're actually checking. I didn't check to see if the code was idiomatic; merely a sanity check.

BurntSushi:
This PR does the following:

* Slightly modulates its phrasing about `time`'s lack of support for the
  time zone database. I've added references to the unofficial (and
  seemingly unmaintained) `time-tz` crate, but its support is quite
  lackluster and comes with all of the same downsides as `chrono-tz`.
* Removes text about `time` being unable to get the local offset from
  UTC in the presence of multiple threads. That seems to have been fixed
  with Rust 2024 (since `std::env::set_var` is now marked `unsafe`, it's
  no longer unsound to call `localtime_r`).
* `time` now has `strftime`, so update the text to include that.
* Refresh the benchmark results. (Not much changes here.)

Fixes #500

### DIFF
--- a/COMPARE.md
+++ b/COMPARE.md
@@ -728,8 +728,7 @@ zoned/fixed_offset_to_civil_datetime                            1.23      5.6±0
 zoned/fixed_offset_to_timestamp                                 3.15      1.2±0.01ns        ? ?/sec    1.00      0.4±0.00ns        ? ?/sec
 ```
 
-Questions about benchmarks are
-welcome in
+Questions about benchmarks are welcome in
 [Discussions on GitHub](https://github.com/BurntSushi/jiff/discussions).
 
 ## [`time`](https://docs.rs/time) (v0.3.36)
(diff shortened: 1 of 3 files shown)


BurntSushi:
@jhpratt happy to make follow-up changes here

jhpratt:
Works for me! Benchmarks aren't a huge worry; once I get more things done I might bother to craft some new ones for various crates.

Issue #373, opened 2025-05-10
a different internal crate organization to help with build times


BurntSushi:
While #372 tries to attack the problem of build times with the obvious "make it possible to compile less stuff," this issue promotes a slightly more radical approach. Namely, it is well known that splitting a crate up into multiple crates will help with compile times, presumably because it helps make more effective use of multiple CPUs.

I did the same thing with `regex`. Although, I didn't do it because of build times. I did it because I wanted to expose lower level primitives outside of a standard regex API (`memchr` and `aho-corasick` being examples of that) and I wanted to expose some of the regex engine internals for more flexible use cases (that's `regex-syntax` and `regex-automata`). I'd guess that this ended up helping build times, although there's no real comparison point there.

My problem with doing this is: 1) it expands the dependency tree and 2) it introduces more semver boundaries. I am _loathe_ to have more semver boundaries. They are a huge fucking pain in the ass. They make releases and refactoring more annoying. Moreover, you have to be super careful about introducing public dependencies, or else the semver API evolution of one crate is tied to another.

With that said... Jiff doesn't actually have a ton of internal code. It has some. But most of what I'd put into separate crates are _already_ in Jiff's public API.

Anyway, here's what I'm thinking:

* `jiff-primitive` would have `Timestamp`, `SignedDuration`, `tz::Offset`, `civil::{DateTime, Date, Time, ISOWeekDate, Weekday}` and maybe some others that I'm thinking about.
* `jiff-tz` would have the TZif and POSIX time zone parsers. It would have a public dependency on `jiff-primitive`. _Probably_ `TimeZone` has to live here too.
* `jiff-fmt` has what's in `jiff::fmt` today, and would have public dependencies on `jiff-primitive` and `jiff-tz`.
* `jiff` glues everything together. It still has some non-trivial code in it, like `Zoned` and `Span`. Along with `TimeZoneDatabase`. But otherwise is a facade over the aforementioned crates. Users of Jiff (like users of `regex`) wouldn't need to worry about the crates above. Jiff's API would stay the same cohesive self. Annoyingly, Jiff would _not_ have public dependencies on any of the above crates.

From my perspective, there are _significant_ challenges with the above blueprint.

[...]

PR #504, opened 2026-02-08
stop using ranged integers internally

### LINKED ISSUE #11: consider ripping out ranged integer types
One of the first abstractions I developed in Jiff was the ranged integer type. It used a very light amount of `const` integers/generics. Instead, its key innovation was that, when `debug_assertions` was enabled, it would keep track of its minimum and maximum possible values, computed via the operations performed on that integer. Each ranged integer would carry with it minimum and maximum values. And so every step of the calculation would be checked not just with respect to the primitive integer representation, but with respect to the bounds set for that particular integer type. For example, `type Year = ri16<-9999, 9999>`.

There are some failings though:

* The ranged integer abstraction seems to inhibit most APIs from being `const`. See #2 for more on that. Although it's worth pointing out that, even if ranged integers were ripped out, there are still potentially significant hurdles from making more of Jiff's APIs `const`. But I think it's safe to say more APIs than today could be feasibly made `const`.
* The ranged integer abstraction increases the use of generics in a variety of places. To make writing code convenient, there are a lot of `number: impl RInto<RangedIntegerType>` parameters on functions. This probably imposes a hit on compilation times and binary size. Some of this could likely be mitigated through explicit monomorphization, but it would be a huge pain.
* Singe ranged integers aren't a publicly exported thing in Jiff (and they shouldn't be), a lot of APIs are split between the internal "ranged integer" API and the normal public "primitive integer" API. This is annoying.
* While in the beginning, my ranged integer types were catching a lot of numerical overflow bugs, I've found that as of recently, most of the overflows they catch are actually due to incorrect use of the ranged integer types rather than legitimate bugs in the code. Namely, the ranged integers types aren't a _perfect_ abstraction. The minimum and maximums aren't always correctly tracked, especially when there's conditional control flow. Still, maybe they are helping prevent bugs just by virtue of providing very consistent range checks on inputs.

So I'm not sure. I'm currently still happy with keeping ranged integers around, but I wonder if they should be looked at more as a tool that helped me "bootstrap" the library and get the numerical algorithms correct. But that they are more like scaffolding, something to be dismantled now that the library's numerical core is pretty much complete.

BurntSushi:
This is a massive refactor that completely deletes Jiff's ranged
integer abstraction. It's replaced with a lighter weight abstraction
for checking ranges and doing basic arithmetic.

The main simplification here is that we no longer need to use ranged
integer types in the representation of datetime or duration types.
This allows for a lot more flexibility in terms of how things are
represented. We also no longer need to carefully manage ranged integers
and their dynamic ranges internally when running conditional code.
Basically, overall, there is a lot less bullshit in the code.

I really loved the idea of ranged integers when I added it. They worked
like Ada ranged integers where computations were actually allowed
to "go outside" of the allowed range temporarily, so long as the
integer came back into its associated range before it was "observed" by
converting it into a primitive integer type. It actually worked really
well to catch boundary bugs in my datetime arithmetic code initially.
But once the code started getting bigger and there was more conditional
logic involving ranged integers, things got supremely annoying. And I
would often run into false positive panics from ranged integers that
were not identifying real bugs but were instead a result of me "holding
it wrong."

They were overall making the code more complicated for not that much
gain. And they were inhibiting changes to representation because of
their "viral" nature. Once you use a ranged integer, you had to keep
using it or else its internal dynamic range was lost.

I was hoping that this would improve compile times by having fewer abstractions
for the compiler to cut through, but instead we only get a modest improvement.
Before:

```
$ hyperfine -w1 --max-runs 3 --prepare 'cargo clean' 'cargo b' 'cargo b -r' ; A bart
Benchmark 1: cargo b
  Time (mean ± σ):      4.717 s ±  0.010 s    [User: 15.928 s, System: 2.202 s]
  Range (min … max):    4.706 s …  4.724 s    3 runs

Benchmark 2: cargo b -r
  Time (mean ± σ):      7.500 s ±  0.043 s    [User: 63.989 s, System: 2.566 s]
  Range (min … max):    7.464 s …  7.547 s    3 runs
```

After:

```
$ hyperfine -w1 --max-runs 3 --prepare 'cargo clean' 'cargo b' 'cargo b -r' ; A bart
Benchmark 1: cargo b
  Time (mean ± σ):      4.407 s ±  0.025 s    [User: 15.425 s, System: 2.175 s]
  Range (min … max):    4.378 s …  4.426 s    3 runs

Benchmark 2: cargo b -r
  Time (mean ± σ):      7.356 s ±  0.029 s    [User: 64.996 s, System: 2.601 s]
  Range (min … max):    7.339 s …  7.390 s    3 runs
```

Moreover, total LLVM lines on my Biff benchmark only decreased by about
2,000.

This also opens the door for potentially experimenting with splitting
Jiff into more crates internally, as discussed in #373. There are still
other issues there---and I'm not sold on doing it---but the status quo
isn't ideal either. (For example, because there are no shared helper
crates, `jiff` and `jiff-static` have copies of the same code. Not all
of Jiff, but not tiny either.) With ranged integers, there was no real
way to split things up without completely negating the benefits of
ranged integers in the first place (or exposing them in the public API,
which seemed nuts).

See #11 for more discussion about ranged integers. I had thought I was
going to live with them, but they were causing too much pain.

In this PR, I also did a lot of refactoring around rounding and durations.
Internally, we use `i128` a lot less (although not zero) in favor of
`SignedDuration`, which is a smaller type with better datetime
integration than just a bare `i128`.

I also removed one use of floating point when rounding calendar
units of a `Span`. This means it's actually plausible to provide a mode
for building Jiff where no `f64` are used at all. The only places where
they are used are in `Span::total` and the `try_from_secs_{f32,f64}`
APIs on `SignedDuration`. I'm not sure if this is something that is
actually desirable though.

Closes #11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ TODO
 
 Enhancements:
 
+* [#11](https://github.com/BurntSushi/jiff/issues/11):
+Stop using ranged integers internally.
 * [#500](https://github.com/BurntSushi/jiff/issues/500):
 Update comparison with the `time` crate in the Jiff documentation.
 * [#502](https://github.com/BurntSushi/jiff/issues/502):


BurntSushi:
Benchmarks generally seem unchanged except for a few noisy ones:

```
$ critcmp baseline update -f jiff -t 10
group                                               baseline                               update
-----                                               --------                               ------
print/iso8601_duration/tiny/duration/buffer/jiff    1.00      7.7±0.05ns        ? ?/sec    1.47     11.4±0.03ns        ? ?/sec
print/iso8601_duration/tiny/span/buffer/jiff        1.00      7.6±0.06ns        ? ?/sec    1.45     11.0±0.02ns        ? ?/sec
print/strftime/oneshot/to_string/jiff               1.29    101.0±0.94ns        ? ?/sec    1.00     78.4±0.72ns        ? ?/sec
zoned/fixed_offset_to_timestamp/jiff                1.00      0.3±0.01ns        ? ?/sec    1.31      0.4±0.00ns        ? ?/sec
```

I would generally expect some operations to have gotten faster, but they aren't yet covered by Jiff benchmarks. In particular, I think in a lot of cases regarding rounding, we're doing less conversion to-and-from 128-bit integers.

PR #506, opened 2026-02-09
tz: add support for `emscripten` system timezone

### LINKED ISSUE #490: Add support for getting system timezone on wasm32-wasi/wasm-emscripten
I've recently implemented wasi & emscripten support for the [iana-time-zone](https://github.com/strawlab/iana-time-zone) crate. Would you like me to add support in Jiff as well? If so should I add a dependency on `iana-time-zone` or copy the implementation?

bschoenmaeckers:
Closes #490

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -514,43 +514,36 @@ jobs:
     - name: Run integration tests
       run: cargo test --test integration -- --nocapture
 
-  # Test the wasm32-unknown-emscripten target.
-  #
-  # Regretably, `insta` doesn't work on emscripten, so we just do a basic
-  # sanity check here.
-  #
-  # At some point, this target just broke and I have no idea how to fix it:
-  # https://github.com/BurntSushi/jiff/actions/runs/21565270214/job/62135646633?pr=492
-  # wasm32-unknown-emscripten:
-  #   runs-on: ubuntu-latest
-  #   env:
-  #     # Bump this as appropriate. We pin to a version to make sure CI
-  #     # continues to work as cross releases in the past have broken things
-  #     # in subtle ways.
-  #     CROSS_VERSION: v0.2.5
-  #   steps:
-  #   - name: Checkout repository
-  #     uses: actions/checkout@v5
-  #   - name: Install Rust
-  #     uses: dtolnay/rust-toolchain@nightly
-  #     with:
-  #       toolchain: stable
-  #   - name: Install and configure Cross
-  #     run: |
-  #       dir="$RUNNER_TEMP/cross-download"
-  #       mkdir "$dir"
-  #       echo "$dir" >> $GITHUB_PATH
-  #       cd "$dir"
-  #       curl -LO "https://github.com/cross-rs/cross/releases/download/$CROSS_VERSION/cross-x86_64-unknown-linux-musl.tar.gz"
-  #       tar xf cross-x86_64-unknown-linux-musl.tar.gz
-  #   - name: Build jiff
-  #     run: cross build --verbose --target wasm32-unknown-emscripten -p jiff
-  #   - name: Build jiff-tzdb
-  #     run: cargo build --verbose --target wasm32-unknown-emscripten -p jiff-tzdb
-  #   - name: Build jiff-tzdb-platform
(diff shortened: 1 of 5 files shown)


src/tz/system/wasm_emscripten.rs
```diff
@@ -0,0 +1,48 @@
+use crate::tz::{TimeZone, TimeZoneDatabase};
+use core::ffi::{c_char, CStr};
+use core::ptr::NonNull;
```
BurntSushi: Can you put the `crate` import below the `core` imports in its own block? (i.e., Make it consistent with the style used throughout Jiff.)

src/tz/system/wasm_emscripten.rs
```diff
@@ -0,0 +1,48 @@
...
+    const SCRIPT: &CStr = {
+        match CStr::from_bytes_with_nul(
+            "Intl.DateTimeFormat().resolvedOptions().timeZone\0".as_bytes(),
+        ) {
+            Ok(s) => s,
+            Err(_) => unreachable!(),
+        }
+    };
+
+    let name = unsafe {
```
BurntSushi: All `unsafe` blocks need a `SAFETY` comment explaining why they are safe.

src/tz/system/wasm_emscripten.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+pub(super) fn get(db: &TimeZoneDatabase) -> Option<TimeZone> {
+    const SCRIPT: &CStr = {
+        match CStr::from_bytes_with_nul(
+            "Intl.DateTimeFormat().resolvedOptions().timeZone\0".as_bytes(),
+        ) {
+            Ok(s) => s,
+            Err(_) => unreachable!(),
+        }
```
BurntSushi: I believe you can use `crate::util::constant::unwrapr!` here.

BurntSushi (changes requested):
Thank you for working on this!

bschoenmaeckers:
Addressed all comments and fixed the CI, so this is ready again for review.

.github/workflows/ci.yml
```diff
@@ -514,43 +514,36 @@ jobs:
...
+      - name: Build jiff
+        run: cargo build --verbose -p jiff
+      - name: Build jiff-tzdb
+        run: cargo build --verbose -p jiff-tzdb
+      - name: Build jiff-tzdb-platform
+        run: cargo build --verbose -p jiff-tzdb-platform
+      - name: Run library tests
+        run: cargo test --lib -- --nocapture
+      - name: Run integration tests
+        run: cargo test --test integration -- --nocapture
```
BurntSushi: Oh wow nice! You got this to work with `insta` too!

src/tz/system/wasm_emscripten.rs
```diff
@@ -0,0 +1,48 @@
+use crate::tz::{TimeZone, TimeZoneDatabase};
+use core::ffi::{c_char, CStr};
+use core::ptr::NonNull;
+
+extern "C" {
+    fn emscripten_run_script_string(script: *const c_char) -> *mut c_char;
```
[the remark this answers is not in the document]
BurntSushi: Hmmm that says:

> Note: If -sDYNAMIC_EXECUTION=0 is set, this function will not be available.

Should we be concerned about that? I guess that means Jiff would fail to build at the linking step. I guess that's fine, and if folks run into a problem they can file an issue. Not sure what we would do instead. I guess we'd probably try to do a `dlopen` or something and check if the symbol is available at runtime.

BurntSushi:
I've fixed up this PR a bit. I split the `unsafe` calls into its component parts and added more trace logging. We should be very careful about whether we use `?` in these system integration points because we swallow errors. I like to have at least a `trace!` whenever something goes wrong so that folks can diagnose problems when and if they get an `Etc/Unknown` time zone.

bschoenmaeckers:
> I've fixed up this PR a bit. I split the `unsafe` calls into its component parts and added more trace logging. We should be very careful about whether we use `?` in these system integration points because we swallow errors. I like to have at least a `trace!` whenever something goes wrong so that folks can diagnose problems when and if they get an `Etc/Unknown` time zone.

Thanks!

PR #509, opened 2026-02-11
jiff-diesel: Allow using Timestamp with diesel::sql_types::Timestamp

teohhanhui:
This should be correct for PostgreSQL, regardless of the timezone set at the database level.

> Timestamps are represented in Postgres as a 64 bit signed integer representing the number of microseconds since January 1st 2000.

https://docs.rs/diesel/latest/diesel/pg/data_types/struct.PgTimestamp.html

(And the same reason why `diesel::sql_types::Timestamp` can be converted to/from `SystemTime` on Postgres.)

### DIFF
--- a/crates/jiff-diesel/src/nullable.rs
+++ b/crates/jiff-diesel/src/nullable.rs
@@ -21,6 +21,10 @@ use crate::ToDiesel;
 )]
 #[cfg_attr(
     feature = "postgres",
+    diesel(sql_type = diesel::sql_types::Nullable<diesel::sql_types::Timestamp>),
+)]
+#[cfg_attr(
+    feature = "postgres",
     diesel(sql_type = diesel::sql_types::Nullable<diesel::sql_types::Timestamptz>),
 )]
 #[cfg_attr(
--- a/crates/jiff-diesel/src/postgres.rs
+++ b/crates/jiff-diesel/src/postgres.rs
@@ -36,6 +36,20 @@ impl ToSql<sql_types::Timestamptz, Pg> for Timestamp {
     }
 }
 
+impl FromSql<sql_types::Timestamp, Pg> for Timestamp {
+    fn from_sql(bytes: PgValue<'_>) -> deserialize::Result<Timestamp> {
+        // The encoding is the number of *microseconds* since
+        // POSTGRES_EPOCH_DATETIME.
+        let PgTimestamp(micros) =
+            FromSql::<sql_types::Timestamp, Pg>::from_sql(bytes)?;
+        let micros = jiff::SignedDuration::from_micros(micros);
+        // OK because the timestamp is known to be valid and in range.
+        let epoch =
+            jiff::Timestamp::from_second(POSTGRES_EPOCH_TIMESTAMP).unwrap();
+        Ok(epoch.checked_add(micros)?.to_diesel())
+    }
+}
+
 impl FromSql<sql_types::Timestamptz, Pg> for Timestamp {
     fn from_sql(bytes: PgValue<'_>) -> deserialize::Result<Timestamp> {
         // The encoding is the number of *microseconds* since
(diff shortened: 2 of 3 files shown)


BurntSushi:
This impl was very specifically and intentionally left out. From the [PostgreSQL documentation](https://www.postgresql.org/docs/current/datatype-datetime.html):

> In a value that has been determined to be timestamp without time zone, PostgreSQL will silently ignore any time zone indication. That is, the resulting value is derived from the date/time fields in the input string, and is not adjusted for time zone.

And:

> Conversions between timestamp without time zone and timestamp with time zone normally assume that the timestamp without time zone value should be taken or given as timezone local time.

In other words, a `TIMESTAMP WITHOUT TIME ZONE` in PostgreSQL is _local_ (i.e., _civil_) time. That's why the `TIMESTAMP WITHOUT TIME ZONE` type maps to Jiff's `civil::DateTime` type and _very specifically and intentionally_ not `jiff::Timestamp`:

https://github.com/BurntSushi/jiff/blob/b303488ae94aeb66500230fe726f778b99f13591/crates/jiff-diesel/src/postgres.rs#L81-L53

The SQL's standard handling of datetime is an important lesson in the idea that "just because someone names it so, does not make it so."

See also: https://stackoverflow.com/questions/5876218/difference-between-timestamps-with-without-time-zone-in-postgresql

> Timestamps are represented in Postgres as a 64 bit signed integer representing the number of microseconds since January 1st 2000.

Yes, that may be how they're _represented_, but that doesn't mean every PostgreSQL `TIMESTAMP WITHOUT TIME ZONE` value is actually a timestamp.

For example, let's take my current local time: `2026-02-11 10:51:00`. Its time in UTC is `2026-02-11 15:51:00` because my current time zone offset is `-05`. That UTC time can be converted to a Unix timestamp. That's `1770825060` seconds since the Unix epoch.

But I _could_ choose to just represent all local datetimes _as if_ they were UTC. For example, by _assuming_ that `2026-02-11 10:51:00` was the time in UTC. Mapping that to a number of seconds since the Unix epoch, I get `1770807060`. Now if I convert that back to an actual UTC time, I get `2026-02-11T10:51:00Z`. The only way I can get my local time is by knowing what that original transformation was, and therefore effectively ignore the `Z` offset and just treat the date and time as local. But if I didn't know about that original transformation, I could just as easily treat it as a real UTC time. And when I convert it back to my local time (remember, `-05` from UTC), I'd get `2026-02-11 05:51:00`. Which was _not_ the local time I started with.

This is an unbelievably subtle footgun, and this is precisely why there will never be an impl like the one proposed in this PR. My example above demonstrates that the _representation_ of a thing and the _meaning_ of a thing are two different things. You can very easily just represent local times _as if_ they were timestamps, but that _doesn't make them actual timestamps_. They don't actually correspond to an instant in time _unless_ you know the offset that was used to get the local time in the first place. But that information is lost when you use `TIMESTAMP WITHOUT TIME ZONE`. In contrast, if you use `TIMESTAMP WITH TIME ZONE`, the offset given is used to convert the datetime to a real timestamp that can always be resolved to a precise instant in time.

> (And the same reason why `diesel::sql_types::Timestamp` can be converted to/from `SystemTime` on Postgres.)

This is an unambiguous mistake in Diesel's API. You won't notice a problem with it if you always use `SystemTime` on the way in and `SystemTime` on the way out. But if the database is using `TIMESTAMP WITHOUT TIME ZONE` to store local times (which, as noted above, is what it's intended use case is), then deserializing those values into a `SystemTime` is virtually guaranteed to produce a nonsense result (unless the original offset was coincidentally `GMT`).

BurntSushi:
I added an example demonstrating this in https://github.com/BurntSushi/jiff/pull/511

I expect this issue to come up again. This way, we have an executable program to point to.

PR #521, opened 2026-02-21
Implement memory_usage() for major jiff types.

### LINKED ISSUE #520: Add (optional) support for typesize
[`typesize`](https://docs.rs/typesize/latest/typesize/) offers a trait to calculate the total memory size used by a type (including heap allocations). This can be important for caches; [PaperCache](https://papercache.io/) uses `typesize` to determine how much memory a particular key or value uses; the Discord library [`serenity`](https://serenity-rs.github.io/serenity/next/serenity/index.html) can optionally use it to track the size of its cached data, and so on.

While `typesize` includes optional implementations for some external libraries like `arrayvec` or `url`, it's currently impossible to fully implement `TypeSize` support for `jiff` from outside the `jiff` crate, [because of how `TimeZone` pointers work](https://github.com/BurntSushi/jiff/blob/0.2.20/src/tz/timezone.rs#L2010-L2032).

I "implemented" support for most of `jiff`'s core types in https://github.com/GnomedDev/typesize/pull/10 (because for a wholly stack-based type the implementation can use default methods):

```rust
impl TypeSize for jiff::SignedDuration {}
impl TypeSize for jiff::Timestamp {}
// etc
```

However, the `typesize` dev [would prefer](https://github.com/GnomedDev/typesize/pull/10#issuecomment-3936492129) to have the support be implemented in the `jiff` crate (which I can understand, since that will make it easier to keep up to date). And of course if we're going to have an implementation at all, it would be nice to support `TimeZone`.

If you're willing to accept a PR for this, I'd be happy to draft one.

inklesspen:
These routines give the heap usage by these types, which can be useful when tracking cache sizes.

I did not implement the routine for formatting/rounding/etc types, since these seem unlikely to be held in caches.

Fixes #520.

### DIFF
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -2016,6 +2016,14 @@ impl Date {
     pub fn series(self, period: Span) -> DateSeries {
         DateSeries { start: self, period, step: 0 }
     }
+
+    /// Returns the heap memory usage, in bytes, of this date.
+    ///
+    /// This does **not** include the stack size used up by this date.
+    /// To compute that, use `std::mem::size_of::<Date>()`.
+    pub fn memory_usage(&self) -> usize {
+        0
+    }
 }
 
 /// Parsing and formatting using a "printf"-style API.
--- a/src/civil/date.rs
+++ b/src/civil/date.rs
@@ -2016,14 +2016,6 @@ impl Date {
     pub fn series(self, period: Span) -> DateSeries {
         DateSeries { start: self, period, step: 0 }
     }
-
-    /// Returns the heap memory usage, in bytes, of this date.
-    ///
-    /// This does **not** include the stack size used up by this date.
-    /// To compute that, use `std::mem::size_of::<Date>()`.
-    pub fn memory_usage(&self) -> usize {
-        0
-    }
 }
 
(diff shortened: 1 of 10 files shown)


BurntSushi:
Thanks! 

I don't think we need to add this for Copy types.

inklesspen:
That's fair!

Also I guess I misunderstood how the `each!` macro would work? I thought `std` would always be available if `alloc` was available. But perhaps it's the other way round.

BurntSushi:
Right. std implies alloc.

inklesspen:
Let me know if you want me to squash down the three commits; otherwise I'll assume the current PR state is fine.

inklesspen:
Is there anything further you need to allow this to be merged?

inklesspen:
This PR is necessary to add jiff compatibility in the serenity library for Discord (https://github.com/serenity-rs/serenity/issues/3484). If there's any way I can help move this forward, please let me know.

BurntSushi (approved):
Looks right to me! Thank you so much and I'm really sorry about the delay here. I appreciate your patience. :-)

BurntSushi:
This PR is on crates.io in `jiff 0.2.24`.

PR #522, opened 2026-02-22
replace serde_yaml 0.9 to yaml_serde 0.10

kazan417:
Now yaml organization forked serde_yaml and officially supports this new yaml_serde.

https://yaml.org/ have link to
https://yaml.com/ on main page link to
https://yaml.com/projects/yaml-serde/
and then it links to https://github.com/yaml/yaml-serde
which is now part of the yaml project https://github.com/yaml/
So we should use this new crate https://crates.io/crates/yaml_serde

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 0.2.21 (TBD)
 ============
-TODO
+
+
+removes the dev-dependency on deprecated `serde_yaml`.
+It has been replaced with `yaml_serde`. See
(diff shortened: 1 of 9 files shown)


BurntSushi:
Since this is a dev-dependency only (and thus such a change really doesn't belong in the CHANGELOG) I'm not inclined to change anything right now. If and when the ecosystem moves to the new crate ([there are only 7 dependents as of 2026-03-03](https://crates.io/crates/yaml_serde/reverse_dependencies)), then I'd be happy to re-evaluate. But I'm not in a rush and I don't think it's worth my time to do the due diligence required to make a change here.

PR #528, opened 2026-03-03
jiff-tzdb: update to tzdb 2026a

BurntSushi:
Ref: https://lists.iana.org/hyperkitty/list/tz-announce@iana.org/thread/ASPLBE3A4BAEXIOQ3KZ6EJSJWBU6L53G/


BurntSushi:
This PR is on crates.io in `jiff 0.2.23`.

PR #535, opened 2026-04-09
Improve comment in Span::checked_add docs

aquarhead:


BurntSushi (approved):
Nice catch! I agree that this is clearer. Thank you. :-)

PR #538, opened 2026-04-17
Replace `windows-sys` with inline bindings

Jake-Shadle:
I noticed when updating crates in one of my projects that `jiff` was pulling in the latest version of windows-sys which was giving me even more duplicates of windows-sys than were already there. I was fairly sure that jiff's usage of windows-sys would be extremely light, which was true, it calls exactly one function and reads one field from one struct. So this PR just replaces windows-sys with the minimal bindings needed by jiff.

Confirmed that it works, at least under wine (I don't have a Windows machine to test on, but presumably there is CI for windows).

```
uptime | wine examples/uptime/target/x86_64-pc-windows-msvc/debug/uptime.exe
[temp-clones/jiff/src/tz/system/windows/mod.rs:110:8] tz_key_name = "Central Europe Standard Time"
2026-03-30T14:39:00+02:00[Europe/Budapest]
```

Looking at #106 there was some interest in doing this. While the suggestion was to use windows-bindgen, I don't think that makes much sense for such a simple use case.

Note that I used `kind = "raw-dylib"` as that doesn't require the use of an import lib, but that required a bump of the MSRV to 1.71 which seems to be ok as the README indicates it will never be a release within the past year, and 1.71 is almost 3 years old at this point.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -329,7 +329,7 @@ jobs:
     - name: Install Rust
       uses: dtolnay/rust-toolchain@master
       with:
-        toolchain: 1.70.0
+        toolchain: 1.71.0
     # We would use `cargo build --all` here, but `jiff-cli` doesn't really
     # track an MSRV and I don't want it to.
     - name: Build jiff
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
     - name: Switch to Rust 2024
       run: sed -i.bak 's/^edition = "2021"$/edition = "2024"/' Cargo.toml
     - name: Bump rust-version to Rust 1.85
-      run: sed -i.bak 's/^rust-version = "1\.70"$/rust-version = "1.85"/' Cargo.toml
+      run: sed -i.bak 's/^rust-version = "1\.71"$/rust-version = "1.85"/' Cargo.toml
     - run: ./scripts/test-default
 
   # Test all features enabled.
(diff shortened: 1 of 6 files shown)


BurntSushi:
@ChrisDenton what do you think about this?

ChrisDenton:
I think it makes sense to me to reduce the windows-sys update churn but I would prefer if it used the [windows-link](https://docs.rs/windows-link/latest/windows_link/) crate for linking imports. This is a very small macro-only crate but it handles the difference between architectures properly and, if someone does need to override linkage of system dlls, it's a single crate to `[patch]`.

We could also generate the bindings using [`windows-bingen`](https://docs.rs/windows-bindgen/latest/windows_bindgen/) but for only a few imports that doesn't seem all that worth it.

BurntSushi:
@Jake-Shadle could you follow @ChrisDenton's advice and use the `windows-link` crate please? Thank you!

Jake-Shadle:
Done.

BurntSushi:
Great thank you. I'll work on merging this PR soon. Today or tomorrow. The msrv bump complicated things somewhat, but it is quite okay to bump to Rust 2024 and make that the msrv. I'll take care of that and then rebase this PR for you. So you shouldn't have any action items here! Thank you.

BurntSushi:
> Great thank you. I'll work on merging this PR soon. Today or tomorrow. The msrv bump complicated things somewhat, but it is quite okay to bump to Rust 2024 and make that the msrv. I'll take care of that and then rebase this PR for you. So you shouldn't have any action items here! Thank you.

This isn't right. The README is unfortunately quite conservative on MSRV bumps pre-1.0. Effectively requiring a semver incompatible release. I don't think it's worth waiting for a `jiff 0.3` or `jiff 1.0` to make this change, especially since it can be done with a Rust 1.71 MSRV. It's a modest bump and technically something I said wouldn't happen, but the README has some grey area around system level dependencies. I'd put this PR in that category.

Anyway, I'll move forward with this. But I'll reserve a Rust 2024 migration for a real semver incompatible release.

BurntSushi:
Actually, the MSRV bump doesn't really make sense to me here. `windows-sys` was already using `windows-link`, so nothing has really changed. It just means it's an artifact of doing business on Windows, which this PR does not change.

BurntSushi:
This is fixed on crates.io in `jiff 0.2.26`.

Issue #34, opened 2024-07-22
write a guide for converting types to and from `chrono` and `time`


BurntSushi:
While it's possible we might want an actual crate to do this, I think we can start by at least documenting what the conversions are and the nuances in those conversions.

bschoenmaeckers:
I'm interested in a `jiff-chrono` crate with conversions like `jiff-icu`. This will make migrating to jiff a lot easier in big projects.

Licht-T:
Working on migrating code from `chrono` to `jiff` surfaced how much of the semantic disagreement in this thread matters at the type level:

- `NaiveDateTime` to `Timestamp` has no well-defined meaning without user input - the naive value could be UTC, a local wall clock in some zone, or a zoneless civil datetime, and chrono's type doesn't say which. Silently picking UTC is a subtle off-by-N-hours bug waiting to ship.
- `DateTime<FixedOffset>` to `Zoned` can only produce a fixed-offset zone, so any calendar arithmetic done afterward will not follow the DST rules of whatever zone the original `chrono` value was *meant* to represent.
- `chrono`'s leap-second encoding has no representation in `jiff`.
- `chrono`'s date and timestamp ranges are wider than `jiff`'s, so all range-narrowing conversions need to be fallible at the type level.

That argues for a dedicated `jiff-chrono` crate that is **strict by default** - custom `ConvertFrom`/`ConvertTryFrom` traits (orphan rule, same pattern as `jiff-icu`), refuses `NaiveDateTime` to `Timestamp` outright so the UTC-vs-local decision is forced to the call site, rejects leap seconds and calendar-unit spans, with a separately-named `lossy` module for opt-in saturation / "days-are-24h" semantics. `DateTime<chrono_tz::Tz>` to `Zoned` behind a feature flag for users who need to preserve IANA zone identity across a round-trip.

I have a working prototype and will open a PR.

avsaase:
I published [`jiff-chrono-conversions`](https://crates.io/crates/jiff-chrono-conversions) with a bunch of conversions between `jiff` and `chrono` types.

PR #540, opened 2026-04-22
jiff-chrono: add integration crate for chrono

Licht-T:
This PR adds a new `jiff-chrono` crate that provides conversions between Jiff types and their counterparts in `chrono` and, optionally, `chrono-tz`. It follows the same structural conventions as `jiff-icu`, `jiff-sqlx`, and `jiff-diesel`: standalone workspace, custom `ConvertFrom` / `ConvertTryFrom` traits (the orphan rule forbids `From` / `TryFrom` between two foreign types), strict-by-default policy, and a separately-named `lossy` module for opt-in semantics.

The covered conversion matrix:

* Civil types (`Date`, `Time`, `DateTime`, `Weekday`) ↔ chrono naive equivalents. Jiff → chrono is infallible; the reverse is fallible due to chrono's wider year range and its leap-second encoding.
* `Timestamp` ↔ `DateTime<Utc>`. The chrono → Jiff side is generic over `Tz: chrono::TimeZone`, so `DateTime<FixedOffset>` and (feature-gated) `DateTime<chrono_tz::Tz>` also convert directly.
* `Zoned` ↔ `DateTime<FixedOffset>`, with a rustdoc caveat that this drops IANA zone identity. Behind a `chrono-tz` feature, `Zoned` ↔ `DateTime<chrono_tz::Tz>` preserves identity.
* `tz::TimeZone` ↔ `chrono_tz::Tz` standalone, also under `chrono-tz`.
* `tz::Offset` ↔ `chrono::FixedOffset`. Jiff → chrono is fallible because Jiff's ±25:59:59 range is wider than chrono's ±23:59:59.
* `SignedDuration` ↔ `TimeDelta`.
* `Span` → `TimeDelta` (strict: rejects any nonzero calendar unit). `TimeDelta` → `Span` and `chrono::Months` → `Span` are infallible.

Some conversions were deliberately omitted:

* `NaiveDateTime → Timestamp` / `NaiveDateTime → Zoned`. A `NaiveDateTime` has no intrinsic meaning — it could be UTC, local, or a zoneless civil reading — and silently picking one is an off-by-N-hours bug waiting to ship. The caller must disambiguate on the chrono side (`.and_utc()` or `.and_local_timezone()`) before converting.
* Anything involving `chrono::Local`.
* `chrono::Days → Span`: as of `chrono` 0.4.44, `chrono::Days` does not expose an accessor for its inner `u64`, so a third-party crate cannot read it.

The `lossy` module provides opt-in helpers for cases where the default conversion refuses a trade-off:

* `naive_time_saturating`, `naive_datetime_saturating`, and `utc_datetime_saturating` clamp chrono leap-second readings (`nanosecond >= 10⁹`) down to `:59.999999999`.
* `span_to_timedelta_days_are_24h` converts a `Span` with nonzero `days` under the civil-day interpretation. Weeks, months, and years still require `Span::to_duration` with an explicit reference.

Features: `std` (default), `alloc`, `chrono-tz`, and `serde` (passthrough; no new serde impls are defined in this crate).

Tests cover boundary cases (range overflow, leap seconds, fixed-offset limits), quickcheck roundtrips, and a feature-combo matrix added to `scripts/test-integrations`.

Ref #34

### DIFF
--- a/crates/jiff-chrono/COPYING
+++ b/crates/jiff-chrono/COPYING
@@ -1,0 +1,3 @@
+This project is dual-licensed under the Unlicense and MIT licenses.
+
+You may use this code under the terms of either license.
--- a/crates/jiff-chrono/Cargo.toml
+++ b/crates/jiff-chrono/Cargo.toml
@@ -1,0 +1,72 @@
+[package]
+name = "jiff-chrono"
+version = "0.1.0"  #:version
+authors = ["Rito Takeuchi <licht-t@outlook.jp>"]
+license = "Unlicense OR MIT"
+homepage = "https://github.com/BurntSushi/jiff/tree/master/crates/jiff-chrono"
+repository = "https://github.com/BurntSushi/jiff"
+documentation = "https://docs.rs/jiff-chrono"
+description = "Conversion routines between Jiff and chrono."
+categories = ["date-and-time"]
+keywords = ["date", "time", "temporal", "zone", "chrono"]
+edition = "2021"
+rust-version = "1.70"
+include = ["/src/*.rs", "COPYING", "LICENSE-MIT", "UNLICENSE"]
+
+# Integration crates in Jiff are explicitly isolated from the workspace to
+# avoid dependencies accumulating. I was originally motivated to do this
+# because rustc kept getting stun-locked compiling diesel. And this was somehow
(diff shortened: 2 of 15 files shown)


BurntSushi:
As #34 says, the starting point here is:

> I think we can start by at least documenting what the conversions are and the nuances in those conversions.

PR #579, opened 2026-05-30
update hifitime to 4.3.0

Devel08:
I was compiling the project and I saw that there was an outdated dependency, when I updated it, a test failed assertion which I fixed too.

### DIFF
--- a/COMPARE.md
+++ b/COMPARE.md
@@ -1153,7 +1153,7 @@ fn main() -> anyhow::Result<()> {
     let e1: Epoch = "2015-06-30T23:00:00 UTC".parse()?;
     let e2: Epoch = "2015-07-01T00:00:00 UTC".parse()?;
     let duration = e2 - e1;
-    assert_eq!(duration, Duration::from_seconds(3_601.0));
+    assert_eq!(duration, Duration::from_seconds(3_600.0));
 
     Ok(())
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -251,7 +251,7 @@ walkdir = "2.5.0"
 
 # hifitime doesn't build on wasm for some reason, so exclude it there.
 [target.'cfg(not(target_family = "wasm"))'.dev-dependencies.hifitime]
-version = "3.9.0"
+version = "4.3.0"
 
 [[test]]
 path = "tests/lib.rs"


Devel08:
Nevermind, didn't look that the test was for leap seconds. Doesn't hifitime > 3.9.0 support leap seconds?

COMPARE.md
```diff
@@ -1153,7 +1153,7 @@ fn main() -> anyhow::Result<()> {
     let e1: Epoch = "2015-06-30T23:00:00 UTC".parse()?;
     let e2: Epoch = "2015-07-01T00:00:00 UTC".parse()?;
     let duration = e2 - e1;
-    assert_eq!(duration, Duration::from_seconds(3_601.0));
+    assert_eq!(duration, Duration::from_seconds(3_600.0));
```
BurntSushi, line 1156: Thanks for updating this, but this test was meant to demonstrate something about leap seconds and durations computed for two instants that cross a leap second. This change is somewhat unexpected, and I wonder if I'm missing something.

cc @ChristopherRabotin do you know what's going on here?

ChristopherRabotin: [Version 4.0 of hifitime](https://github.com/nyx-space/hifitime/releases/tag/4.0.0), the one that safely landed Firefly Blue Ghost on the Moon (first US Moon lander since Apollo), introduced some big changes, notably keeping track of the initialization time scale and reporting time deltas on the same time scale as both, or in the one without leap seconds, because it's much more intuitive. Your original test case can be reproduced by converting one of the two time scales to TAI (or another time scale which does not have leap seconds).

Devel08: @BurntSushi Can I refactor the test by converting e2 to TAI and comparing the difference between e2 and e1 with Duration::from_seconds(3_601.0)? I already did it and it passes the test with the leap second, I'll just have to open another PR.

Devel08: Also, thanks @ChristopherRabotin!

BurntSushi: Yeah but the surrounding wording should be changed. Probably to include a paraphrasing of what @ChristopherRabotin said 

You can just update this PR. No need for another.

Devel08: Unfortunately, I deleted the repo that this PR was made from, I opened a new one with added detail in the description of the hifitime test.

PR #580, opened 2026-05-30
update hifitime to 4.3.0

Devel08:
since hifitime 4.0 the leaps second test would stop working unless one of the utc time scales isn't converted to TAI and then the difference between them is compared.

### DIFF
--- a/COMPARE.md
+++ b/COMPARE.md
@@ -1144,14 +1144,18 @@ data type in `hifitime` for representing civil time.
 
 In particular, when computing a duration from two `Epoch` values that spans
 a positive leap second (a second gets repeated), `hifitime` will correctly
-report the accurate duration:
+report the accurate duration. With that said, to get an accurate duration, one
+of the two time scales has to be converted to to TAI (or another time scale
+which does not have leap seconds):
 
 ```rust
-use hifitime::{Duration, Epoch};
+use hifitime::{Duration, Epoch, TimeScale};
 
 fn main() -> anyhow::Result<()> {
     let e1: Epoch = "2015-06-30T23:00:00 UTC".parse()?;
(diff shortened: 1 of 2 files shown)


Devel08:
@BurntSushi Sorry for pinging, what do you think?

COMPARE.md
```diff
@@ -1144,14 +1144,18 @@ data type in `hifitime` for representing civil time.
 
 In particular, when computing a duration from two `Epoch` values that spans
 a positive leap second (a second gets repeated), `hifitime` will correctly
-report the accurate duration:
+report the accurate duration. one of the two time scales has to be converted to to TAI (or another time scale which does not have leap seconds):
```
BurntSushi:
```suggestion
report the accurate duration. With that said, to get an accurate duration, one
of the two time scales has to be converted to to TAI (or another time scale
which does not have leap seconds):
```

COMPARE.md
```diff
@@ -1144,14 +1144,18 @@ data type in `hifitime` for representing civil time.
...
+use hifitime::TimeScale;
 use hifitime::{Duration, Epoch};
 
 fn main() -> anyhow::Result<()> {
     let e1: Epoch = "2015-06-30T23:00:00 UTC".parse()?;
-    let e2: Epoch = "2015-07-01T00:00:00 UTC".parse()?;
+    let e2utc: Epoch = "2015-07-01T00:00:00 UTC".parse()?;
+
+    let e2 = e2utc.to_time_scale(TimeScale::TAI);
+
```
BurntSushi:
```suggestion
    let e2: Epoch = "2015-07-01T00:00:00 UTC".parse()?;
    let e2 = e2.to_time_scale(TimeScale::TAI);

```

PR #581, opened 2026-05-31
Add `defmt` support to `jiff` (try again)

dickermoshe:
This PR introduces support for `defmt` for `jiff`. 

In this implementation, `Debug` and `defmt::Format` have the identical outputs, a single notable exception.
Logging `PosixTimeZone` in `defmt` will log `PosixTimeZone(N/A)` instead of properly formatting the contents. Logging this internal state is rarely useful, so this is a worthwhile trade-off. When logging with defmt, we must not use any of the `core::fmt` or we will be pulling in the heavy formatting code, which embedded targets try to avoid. 

Only types which are public (and their dependencies) have `defmt::Format` derived.  
We did not simply derive `defmt::Format` on every type in the repo. 

The approach taken here has a downside. Typically, when logging a type, `defmt` we'll try to optimize how this is sent over the wire using some fancy tricks. However, our implementation has all the formatting done on the device. We are only logging strings. It is quite possible that there is no difference between using `defmt::Display2Format` and `defmt::Debug2Format`, other than the fact that we are no longer pulling in `core:fmt` for `PosixTimeZone(...)`


dickermoshe:
I just realized that this will need to be dealt with, or we are still pulling in `core::fmt`

https://github.com/BurntSushi/jiff/blob/66f6ec2d0b75112deadf7e067dbf2ef114d41c50/src/fmt/temporal/printer.rs#L476-L488

dickermoshe:
>  But... who gives a fuck about printing POSIX time zone strings?

I very much agree. 2 options here
1. Write a printing API for PosixTimeZone anyway, `Format` will match `Display`.
2. Formatting `PosixTimeZone` with look ugly as sin (`PosixTimeZone { some date: MoreInternalState { ahhhh: true}}`).

Both of these options will successfully remove `core:fmt` as a dependency when using defmt.

dickermoshe:
@BurntSushi never mind. fresh eyes showed some issues with this. don' review just yet. I'll ping you when ready

dickermoshe:
@BurntSushi 
You need to make a printer for `PosixTimeZone` first. core::fmt is still in the loop.

BurntSushi:
I would say don't block on me right now and don't worry too much about printing posix time zones. It's an extremely odd case, although it does seem more likely to come up in an embedded environment. In any case, just print something silly for now, like a static `PosixTimeZone(N/A)` or something like that.

dickermoshe:
There are 2 reasons `defmt` is useful.

1. When logging certain types, it is able to optimize the size of the logs. For instance, if it knows it's logging a `u8`, it is able to send over that number in less memory. 
2. Formatting certain types to strings is surprisingly complicated. The code for this is very long and bloats the binary. 

Right now, we are struggling with how to format the custom `Debug/Display impls`
When not using `defmt`, `jiff` will log this to a buffer before and then `write!` it out. 

If we were to do the same thing but with `defmt::write!`, we wouldn't really be getting any of the benefits of `defmt`. 
The formatting code which `jiff` has is already really good. It does not use any of the heavy code in `core::fmt` which normally bloats the binary. The only exception for this was `PosixTimeZone` which we worked around.
Additionally, we are not getting any of the benefits of more compact lugs because logging strings it's not something that `defmt` is able to optimize. 

This PR is ready for review, but after noting the above, it may be worth it either 
1. Making much larger changes to `crate::fmt`, logging types like numbers will use a different path for `defmt`. Instead of turning numbers into strings and then logging as a string, we log the numbers directly. 
2. Logging the types directly. Even though this will be ugly to read, it will still be much more compact. 
3. Keeping this PR as is and enjoy the benefits we are getting for the public types that we `#derive(Format)` 



(You better not think that this was AI. This took me like four tries to write until it made sense to me and was easy to read. 😉)

BurntSushi:
Thank you for the write-up! I apologize for what AI is doing to our discourse here. Ugh. I'm still trying to grapple with it myself.

> Making much larger changes to crate::fmt, logging types like numbers will use a different path for defmt. Instead of turning numbers into strings and then logging as a string, we log the numbers directly.

Hmmm. Interesting. This may not be a terribly big change, but yeah, definitely a deeper one than what you are trying to do.

I think I might need to play with this myself to offer concrete advice.

Can we just write strings for now even if the impl is imperfect?

dickermoshe:
>Can we just write strings for now even if the impl is imperfect?

Yeah, it's fine for now IMHO.

btw, I cam across this today and it is fantastic, just wanted to share it with you

https://diffshub.com/BurntSushi/jiff/pull/581

dickermoshe:
@BurntSushi I don't mean to be a dick, but do you have any sorta ETA for reviewing this?
I think all concerns were addressed, unless I misunderstood something.

Thanks

BurntSushi:
I'm honestly not sure. This week has been crazy, in part because of [this](https://burntsushi.net/encephalitis) and next week I'm on vacation. I'll see what I can do in the next few days, but right now I'm working on a higher priority bug in `regex` so it's possible this is going to take some time.

scripts/test-all
```diff
@@ -7,4 +7,20 @@ set -e
 cd "$(dirname "$0")/.."
 
 echo "===== ALL FEATURES ====="
-cargo test --all-features
```
BurntSushi, line 10: Hmm this isn't great. Why can't we include defmt testing here?

scripts/test-various-feature-combos
```diff
@@ -25,3 +25,15 @@ for f in "${features[@]}"; do
...
+defmt_features=(
+    "defmt"
+    "alloc defmt"
+    "std defmt"
+    "alloc defmt serde"
+    "std defmt serde"
+)
+for f in "${defmt_features[@]}"; do
+    echo "===== CHECK FEATURES: '$f' ====="
+    cargo check --no-default-features --features "$f"
```
BurntSushi: Why can't we run tests here?

src/civil/date.rs
```diff
@@ -2148,6 +2148,15 @@ impl core::fmt::Display for Date {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for Date {
+    fn format(&self, f: defmt::Formatter) {
+        use crate::fmt::DefmtWrite;
+
+        let _ = DEFAULT_DATETIME_PRINTER.print_date(self, DefmtWrite(f));
```
BurntSushi: This should have an unwrap or an expect. We don't want to silently swallow an error here.

src/civil/date.rs
```diff
@@ -3471,6 +3485,7 @@ enum DateWithYear {
 /// Note that when "day of year" or "day of year no leap" are used, then if a
 /// day of month is set, it is ignored.
 #[derive(Clone, Copy, Debug)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
```
BurntSushi: More broadly: can we drop these less important impls and just stick to the principal jiff types? I think they are: SignedDuration, Span, Timestamp, Zoned, tz::TimeZone, civil::{Date, DateTime, Time, Weekday, ISOWeekDay} and Error.

src/error/mod.rs
```diff
@@ -643,6 +668,15 @@ impl core::fmt::Debug for IOError {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for IOError {
+    fn format(&self, f: defmt::Formatter) {
+        // `std::io::Error` does not implement `defmt::Format`. Since this error is std-only
+        // and defmt is mainly used in embedded contexts, omitting the error is fine.
```
BurntSushi: Wrap comments to 79 columns (inclusive) please. :)

src/error/mod.rs
```diff
@@ -643,6 +668,15 @@ impl core::fmt::Debug for IOError {
...
+        defmt::write!(f, "IOError");
```
BurntSushi:
```suggestion
        defmt::write!(f, "IOError(unavailable)");
```

src/error/mod.rs
```diff
@@ -685,6 +719,16 @@ impl core::fmt::Debug for FilePathError {
     }
 }
 
+#[cfg(feature = "defmt")]
+impl defmt::Format for FilePathError {
+    fn format(&self, f: defmt::Formatter) {
+        // `std::path::PathBuf` does not implement `defmt::Format`.
+        // Since this error is std-only and defmt is mainly used in
+        // embedded contexts, omitting the path is fine.
+        defmt::write!(f, "FilePathError");
```
BurntSushi:
```suggestion
        defmt::write!(f, "FilePathError(unavailable)");
```

src/error/util.rs
```diff
@@ -178,3 +181,12 @@ impl core::fmt::Display for OsStrUtf8Error {
         }
     }
 }
+
+#[cfg(feature = "defmt")]
+impl defmt::Format for OsStrUtf8Error {
+    fn format(&self, f: defmt::Formatter) {
+        // `OsStr` does not implement `defmt::Format`. Since this error is std-only
+        // and defmt is mainly used in embedded contexts, omitting the value is fine.
+        defmt::write!(f, "OsStrUtf8Error");
```
BurntSushi:
```suggestion
        defmt::write!(f, "OsStrUtf8Error(unavailable)");
```

src/fmt/mod.rs
```diff
@@ -468,3 +477,39 @@ impl<W: Write> core::fmt::Write for StdFmtWrite<W> {
         self.0.write_str(string).map_err(|_| core::fmt::Error)
     }
 }
+
+/// An adapter for writing Jiff's formatted output to a `defmt` formatter.
+///
+/// This is used by Jiff's `defmt::Format` impls to share the same printer code
+/// used for `core::fmt` output.
```
BurntSushi: This should have a doc example.

src/fmt/temporal/printer.rs
```diff
@@ -474,6 +475,10 @@ impl DateTimePrinter {
             return wtr.write_str(bbuf.filled());
         }
         if let Some(posix_tz) = tz.posix_tz() {
+            // POSIX rule text is not worth pulling core::fmt into defmt logs.
+            if wtr.is_defmt() {
```
BurntSushi: Can we instead derive format on posix time zone with the same string? Otherwise having a new public method on a public trait for just this case is not what I would want. Especially if it is temporary.

BurntSushi (changes requested):

scripts/test-all
dickermoshe: You cannot even compile `defmt` for x86 without getting a linking error.
If we want to test this, we will need to use QEMU to run this on risc-v

src/civil/date.rs
dickermoshe: The issue is that the `jiff::Error` will contain a lot of these internal types, and it really goes all the way down to the bottom.

src/civil/date.rs
dickermoshe: The alternative to silently swallowing the error is panicking, I would hate for my microcontroller to outright crash because we couldn't print a log to the screen.

dickermoshe: I'm reading now online that print itself also will panic if formatting fails. So I guess this is reasonable

src/fmt/temporal/printer.rs
dickermoshe: I made modifications, so now by default, logging any of the internal types will not pull in `core::fmt`.

It is technically possible that if someone were to do the following manually

```rust
DateTimePrinter::new().print_time_zone(&tz, DefmtWrite(f))
```
It would pull it in `core::fmt`, however, if someone wants to stab himself in the eyes that badly, I think we should just let him.

dickermoshe: Logging a Posix time zone will return a `Posix(N/A)`

scripts/test-all
dickermoshe: I'll take a swing at this, adding it to a get-up action.
Prepare for 87 failed github actions

dickermoshe: I actually got it working surprisingly easily. 
Now, how do you want me to have the scripts? 
`defmt` still cannot be run on x86. 

Do you want two scripts:
1. One called "test-all" and one called "test-all-defmt"
3. Should the script check the current CPU architecture and decide what to run at runtime?

src/civil/date.rs
BurntSushi: Yes. Jiff will happily panic in a variety of places because of a bug. This is nothing new.

See https://burntsushi.net/unwrap

dickermoshe:
I believe I addressed all your concerns and need feedback on a couple of items.
Thank you so much for your prompt review. I really appreciate it.

src/civil/date.rs
dickermoshe: I believe in embedded programming circles, they look at panicking in a worse way. 
In this instance, it's just logging, you wouldn't necessarily want some critical piece of hardware to just panic for something relatively minor. 

However, embedded Rust is much more niche, (although increasingly less so).

dickermoshe: I have read that article before, and I didn't know it was from you.

BurntSushi: Basically an unwrap here is no different than `slice[i]`. Jiff has panicking branches literally everywhere. :-)

scripts/test-all
BurntSushi: I am very confused. On this PR, I can do `cargo t --features defmt`. It works and all tests pass. I'm on x86-64.

src/civil/date.rs
BurntSushi: OK, I'll have to go through these myself then.

scripts/test-various-feature-combos
BurntSushi: The answer to this question is that running tests requires `std` in some fashion. Which the thumb target does not support. Even when you do `--no-default-features` (i.e., no std or alloc), the tests still enable `std` and `alloc`, and thus testing there fails.

Running tests without any std or alloc is pretty hard I think, and I've generally punted on that.

But we don't need qemu just to build with `defmt`. So we shouldn't need the "list all features except for `defmt`" stuff either. e.g., this works just fine: `cargo t --features defmt`.

I pushed an update to the CI config that just uses `cargo build`.

scripts/test-all
BurntSushi: I reverted this.

BurntSushi:
I'm going to take over this PR if you don't mind. That just means I'm going to force push to it and squash things down to 1 commit.

dickermoshe:
> I'm going to take over this PR if you don't mind.

Thanks, I don't mind. Let me know if need me again.

BurntSushi:
OK, I've cleaned this up quite a bit and paired down the set of types that implement `defmt::Format`. We mostly just stick to the core primitives and the error types.

I haven't fixed the POSIX time zone problem yet though. I'm inclined to punt on that and treat it as future work, or does that just completely negate the point of using `defmt` in the first place?

Other than that, I think this is ready to be merged and released.

BurntSushi:
@dickermoshe

dickermoshe:
LGTM

Looks like some changes unrelated to defmt were introduced in `src/fmt/temporal/parser.rs`, but they may be related somehow

BurntSushi:
> LGTM
> 
> Looks like some changes unrelated to defmt were introduced in `src/fmt/temporal/parser.rs`, but they may be related somehow

A commit by commit review will shine a light. :)

BurntSushi:
I was able to rejigger the code so that we do indeed completely avoid `core::fmt::Write` now. I basically just moved the `Display` impls for `PosixTimeZone` (and its constituent types) to routines on the temporal printer. That matches what we do with basically everything else, and is generally how we avoid `core::fmt::Write`. _Thankfully_ that was pretty easy. It ended up being a nice side effect of Jiff's internal formatting architecture to be written in a way that doesn't depend on `core::fmt::Write`. Phew. It's a seeming coincidence because the [original motivation for `jiff::fmt::Write`](https://docs.rs/jiff/latest/jiff/fmt/trait.Write.html#design) doesn't have anything to say about core-only use cases.

BurntSushi (approved):

PR #584, opened 2026-06-13
Make `Zoned::DEFAULT` `pub` so one can use it in `const` contexts

xkr47:
`Zoned::default()` isn't `const` and can't be..

### DIFF
--- a/src/zoned.rs
+++ b/src/zoned.rs
@@ -114,6 +114,13 @@ use crate::{
 /// For more information on the specific format supported, see the
 /// [`fmt::temporal`](crate::fmt::temporal) module documentation.
 ///
+/// # Default value
+///
+/// For convenience, this type implements the `Default` trait. Its default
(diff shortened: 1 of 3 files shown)


BurntSushi:
I think this is probably okay, but:

1. We don't do it for other types.
2. The docs need to be written for this.

Out of curiosity, what is your use case for this?

src/zoned.rs
```diff
@@ -385,7 +385,7 @@ impl Zoned {
     /// `Zoned::new` call, but it looks like it did. However, this is a
     /// more robust way to guarantee that `Zoned::default()` is always
     /// fast.
-    const DEFAULT: Zoned = Zoned {
+    pub const DEFAULT: Zoned = Zoned {
```
BurntSushi: I also think this should probably be named `UNIX_EPOCH`. That would make it consistent with `Timestamp`.

BurntSushi (changes requested):

BurntSushi (approved):

PR #597, opened 2026-07-05
tz: return bundled tzdb from db() when std is unavailable

### LINKED ISSUE #533: no_std + tzdb panic
I am trying to use jiff with the following features:
```  
jiff = { version = "0.2.23", default-features = false, features = [
  "alloc",
  "perf-inline",
  "serde",
  "tzdb-bundle-always"
] }
```

The intention behind this is that I do not want jiff to do any IO (read files or environment variables), and I require handling of timezones, therefore I am bundling the tzdb and implicitly setting `no_std`.
I am not allowed / do not want to do unnecessary syscalls.

Is this the correct way to do this? 
currently, this panics with:
```
thread 'main' (35272228) panicked at src/main.rs:161:30:
called `Result::unwrap()` on an `Err` value: failed to find time zone `America/New_York`, since there is no global time zone database configured (and is currently impossible to do so without Jiff's `std` feature enabled, if you need this functionality, please file an issue on Jiff's tracker with your use case)
```

(this is when constructing any civil time, such as:

```rust
                    // clock: jiff::civil::date(2024, 7, 4)
                    //     .at(12, 0, 0, 0)
                    //     .in_tz("America/New_York")
                    //     .unwrap(),
```
)

Thank you.

momomuchu:
Without `std`, `db()` returned an empty database even when the tzdb was
compiled in via `tzdb-bundle-always`/`tzdb-bundle-platform`, so `in_tz`
lookups failed in that configuration. `tzdb-zoneinfo` and
`tzdb-concatenated` both require `std`, so the bundled database is the
only source available there. `db()` now returns it directly.

To keep the bundled path allocation-free, the bundled database is stored
outside the `Arc`. It owns no data of its own (the tzdb is embedded and
parsed zones are cached in a global), so it needs no allocation and can
be built in a `const`. The tradeoff is that `TimeZoneDatabase` grows to
two words, since the internal repr now needs a tag to distinguish the
empty, bundled and `Arc`-backed variants. The size test's comment already
allows this "with good reason", but I'm happy to look at a one-word
variant if you'd prefer to keep it at one.

Fixes #533

### DIFF
--- a/src/tz/db/mod.rs
+++ b/src/tz/db/mod.rs
@@ -32,8 +32,33 @@ pub fn db() -> &'static TimeZoneDatabase {
     // #[cfg(any(not(feature = "std"), miri))]
     #[cfg(not(feature = "std"))]
     {
-        static NONE: TimeZoneDatabase = TimeZoneDatabase::none();
-        &NONE
+        // Without `std` there's no lazily initialized global state. But when
+        // the tzdb is bundled into the binary, we can still hand out a usable
+        // database: the bundled database needs no allocation, so it can be
+        // constructed in a `const`. (Without `std`, the zoneinfo and
+        // concatenated databases are both unavailable, so there's nothing to
+        // prefer over the bundle.)
+        //
+        // Ref: https://github.com/BurntSushi/jiff/issues/533
+        #[cfg(any(
+            feature = "tzdb-bundle-always",
+            all(
+                feature = "tzdb-bundle-platform",
+                any(windows, target_family = "wasm"),
+            ),
+        ))]
+        static DB: TimeZoneDatabase = TimeZoneDatabase {
+            inner: Repr::Bundled(bundled::Database::new()),
+        };
+        #[cfg(not(any(
+            feature = "tzdb-bundle-always",
+            all(
+                feature = "tzdb-bundle-platform",
+                any(windows, target_family = "wasm"),
+            ),
+        )))]
+        static DB: TimeZoneDatabase = TimeZoneDatabase::none();
+        &DB
     }
     // #[cfg(all(feature = "std", not(miri)))]
     #[cfg(feature = "std")]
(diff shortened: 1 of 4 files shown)


src/tz/db/mod.rs
```diff
@@ -714,8 +753,38 @@ fn special_time_zone(name: &str) -> Option<TimeZone> {
...
     use super::*;
 
-    /// This tests that the size of a time zone database is kept at a single
-    /// word.
+    /// Tests that the global database returns the bundled tzdb when the bundle
+    /// is enabled but neither the zoneinfo nor concatenated databases are.
+    ///
+    /// This configuration doesn't have `std` (since both `tzdb-zoneinfo` and
+    /// `tzdb-concatenated` imply it), and so without special handling `db()`
+    /// returns an empty database despite the tzdb being compiled in.
```
BurntSushi:
```suggestion
    /// used to return an empty database despite the tzdb being compiled in.
```

src/tz/db/mod.rs
```diff
@@ -725,7 +794,7 @@ mod tests {
         #[cfg(feature = "alloc")]
         {
             let word = core::mem::size_of::<usize>();
-            assert_eq!(word, core::mem::size_of::<TimeZoneDatabase>());
+            assert_eq!(2 * word, core::mem::size_of::<TimeZoneDatabase>());
```
BurntSushi, line 797: Yeah this seems fine. Going lower than this probably requires pointer tagging, and that doesn't seem worth it. Especially since most uses are of a `&'static TimeZoneDatabase` anyway.

BurntSushi (changes requested):
Thank you! Lovely comments.

Could you split this into two commits please? One that doe the internal representation refactor and then one with the actual change to support the bundled database in `alloc`. The regression test should go in the second commit.

BurntSushi (approved):
Nice, thank you!

PR #609, opened 2026-07-19
cargo: switch to a virtual manifest

BurntSushi:
RIP good revision history


Alc-Alc:
@BurntSushi apologies for the ping. Just for my understanding, do you mind explaining your PR description a bit more?

Is the concern that there will be (well, "is" at this point) a `cargo: switch to a virtual manifest` commit scattered across various files / folders? If so, would doing something like https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view help?

Once again, apologies for the ping and for the question.

PR #594, opened 2026-06-29
Use `crate::…` instead of `jiff::…` in a couple of lib tests

musicinmybrain:
Without this change, these two tests, added in 0.2.30, fail to build downstream in Fedora’s `rust-jiff` package:

```
error[E0433]: cannot find module or crate `jiff` in this scope
    --> src/fmt/buffer.rs:1401:13
     |
1401 |         use jiff::{civil::date, fmt::temporal::DateTimePrinter};
     |             ^^^^ use of unresolved module or unlinked crate `jiff`
     |
     = help: if you wanted to use a crate named `jiff`, use `cargo add jiff` to add it to your `Cargo.toml`

error[E0433]: cannot find module or crate `jiff` in this scope
    --> src/fmt/buffer.rs:1421:13
     |
1421 |         use jiff::{civil::date, fmt::temporal::DateTimePrinter};
     |             ^^^^ use of unresolved module or unlinked crate `jiff`
     |
     = help: if you wanted to use a crate named `jiff`, use `cargo add jiff` to add it to your `Cargo.toml`

For more information about this error, try `rustc --explain E0433`.
```

PR #611, opened 2026-07-19
Fixes for testing from released crates since jiff-core split

musicinmybrain:
Similarly to https://github.com/BurntSushi/jiff/pull/594, this lets me run the tests downstream in Fedora’s `rust-jiff` package, which is based on the release `.crate` archive from crates.io (and also uses integration tests and test data files from the GitHub archive).

I think I got the details right here. It needs to be `crate::civil::DateTime` rather than `jcore::civil::DateTime` because it needs to impl `Display`.

### DIFF
--- a/crates/jiff/src/tz/tzif.rs
+++ b/crates/jiff/src/tz/tzif.rs
@@ -43,8 +43,8 @@ mod tests {
             }
         }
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Ah yup, thank you!

musicinmybrain:
Appeased the rustfmt check and force-pushed.

PR #616, opened 2026-07-26
Remove redundant cast

FranciscoTGouveia:
I believe this cast can be safely removed, as the source type is already the target type.

For reference, this change was found by an optimizer developed as a research project.

CC @nunoplopes

### DIFF
--- a/crates/jiff/src/fmt/util.rs
+++ b/crates/jiff/src/fmt/util.rs
@@ -761,7 +761,6 @@ pub(crate) fn parse_temporal_fraction<'i>(
         // OK because parsing is forcefully limited to 9 digits,
         // which can never be greater than `999_999_99`,
         // which is less than `u32::MAX`.
-        let nanoseconds = nanoseconds as u32;
         Ok(Parsed { value: Some(nanoseconds), input })
     }
 


crates/jiff/src/fmt/util.rs
```diff
@@ -761,7 +761,7 @@ pub(crate) fn parse_temporal_fraction<'i>(
         // OK because parsing is forcefully limited to 9 digits,
         // which can never be greater than `999_999_99`,
         // which is less than `u32::MAX`.
-        let nanoseconds = nanoseconds as u32;
+        let nanoseconds = nanoseconds;
```
FranciscoTGouveia: If I am not overlooking anything, I believe this let-binding can actually be removed.

BurntSushi: Yes. There was probably a time when `parse::fraction` returned an `i32` instead of a `u32`. So it can just be removed.

BurntSushi (changes requested):

crates/jiff/src/fmt/util.rs
FranciscoTGouveia: Done!

BurntSushi (approved):

PR #617, opened 2026-07-28
core: fix time zone offset subtraction

jif-oai:
Fix `jiff_core::tz::Offset::checked_sub`, which incorrectly added seconds instead of subtracting them + add regression tests for signed inputs, overflow, and the `-` / `-=` operators`


BurntSushi:
Thank you!

PR #620, opened 2026-08-02
build(jiff-static): upgrade syn to v3

tisonkun:
cc @BurntSushi 

`syn` v3 requires MSRV >= 1.71 while `jiff` is now 1.70. Perhaps upgrade the MSRV.

### DIFF
--- a/crates/jiff-static/Cargo.toml
+++ b/crates/jiff-static/Cargo.toml
@@ -40,4 +40,4 @@ jcore = { package = "jiff-core", version = "0.1", path = "../jiff-core" }
 jiff-tzdb = { version = "0.1.6", path = "../jiff-tzdb", optional = true }
 proc-macro2 = "1.0.93"
 quote = "1.0.38"
-syn = "3.0.3"
+syn = ">= 2.0.98, < 4"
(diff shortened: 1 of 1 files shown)


tisonkun:
The MSRV CI job passed because it doesn't build `jiff-static`.

BurntSushi:
I'm going to roll this into the 1.0 release because of the MSRV bump. I don't normally consider them breaking changes, but I foolishly promised I'd only do MSRV bumps in minor version changes, _specifically_ even during pre-1.0.

tisonkun:
@BurntSushi Got it. Then I may close this PR as postponed. Since it's a one-line dependency bump you should be able to handle it when release 1.0.

What do you think? BTW, is there an estimation to 1.0?

crates/jiff-static/Cargo.toml
```diff
@@ -40,4 +40,4 @@ jcore = { package = "jiff-core", version = "0.1", path = "../jiff-core" }
 jiff-tzdb = { version = "0.1.6", path = "../jiff-tzdb", optional = true }
 proc-macro2 = "1.0.93"
 quote = "1.0.38"
-syn = "2.0.98"
+syn = "3.0.3"
```
AudaciousAxiom: @BurntSushi It seems that it would be possible to allow both `syn` v2 and `syn` v3 (all tests seem to pass after forcing `jiff-static` to use `syn` v3 in the lockfile):

```suggestion
syn = ">= 2.0.98, < 4"
```

I don't think that would increase `jiff-static`'s MSRV (as a compatible version of `syn` v2 could still be selected by dependents), while also allowing for dependency deduplication for dependents that are not as MSRV-constraint.

tisonkun: Looks like an interesting workaround. I've merged your patch and let's see what @BurntSushi says :D

tisonkun:
Seems would be included in https://github.com/BurntSushi/jiff/pull/622.

Closing ...

PR #628, opened 2026-08-05
Fix fmt::serde serializing Option::Some

### LINKED ISSUE #627: fmt::serde custom serializers do not serialize Option::Some
For JSON this is not a problem, but for binary serializers (like `postcard`) it is. Not serializing Option::Some causes lose of information and serialized data cannot be then deserialized back.

martin-kolarik:
Fixes #627.

### DIFF
--- a/crates/jiff/src/fmt/serde.rs
+++ b/crates/jiff/src/fmt/serde.rs
@@ -248,6 +248,15 @@ pub mod duration {
                 }
             }
 
+            impl<'a> serde_core::Serialize for CompactDuration<'a> {
+                fn serialize<S: serde_core::Serializer>(
+                    &self,
+                    se: S,
+                ) -> Result<S::Ok, S::Error> {
+                    se.collect_str(self)
+                }
+            }
+
             /// Serialize a required `SignedDuration` in the [`friendly`]
             /// duration format using compact designators.
             #[inline]
@@ -267,7 +276,9 @@ pub mod duration {
             ) -> Result<S::Ok, S::Error> {
                 match *duration {
                     None => se.serialize_none(),
-                    Some(ref duration) => required(duration, se),
+                    Some(ref duration) => {
+                        se.serialize_some(&CompactDuration(duration))
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


crates/jiff/src/fmt/serde.rs
```diff
@@ -411,7 +417,9 @@ pub mod span {
             ) -> Result<S::Ok, S::Error> {
                 match *span {
                     None => se.serialize_none(),
-                    Some(ref span) => required(span, se),
+                    Some(ref span) => {
+                        se.serialize_some(&CompactSpan(span).to_string())
```
BurntSushi: I don't think we should be allocating here.

BurntSushi (changes requested):
Thank you. I agree this is a bug. But the allocations here seem unfortunate. Can we remove them? I think it's probably a requirement as these serde impls themselves were specifically written not to require allocation, and Jiff supports enabling serde without enabling `alloc`. It's probably a failing of CI that it isn't failing on this PR. I'm kind of surprised by this actually, since I think this CI job should catch it:

https://github.com/BurntSushi/jiff/blob/84221585adc807156e7ca6990166305d11f494dc/.github/workflows/ci.yml#L133-L153

crates/jiff/src/fmt/serde.rs
BurntSushi: I think we need to implement `Serialize` on `CompactSpan` (and similar types), and that impl should call `collect_str`. Then we can pass `CompactSpan` directly to `se.serialize_some`?

martin-kolarik: Hi, yes. I saw the allocation in the diff after pushing the PR and I was surprised. But it is obvious, that `to_string` requires `alloc`. Implementing `Serialize` on the newtypes definitely would help. On the other side, `collect_str` is already behind `alloc` (or `std`) feature gate so all code calling `collect_str` require `alloc`. Maybe `alloc` is unavoidable in principle and which way to select -- admitting `alloc` import or implementing `Serialize` -- is more a design decision?

I can easily implement `Serialize`/`Deserialize`. It looks like better solution than the current, IMO.

martin-kolarik: I updated the PR, `impl Serialize` is there.

BurntSushi: Using `collect_str` is correct here because it's still an API item even when `serde/alloc` isn't enabled. In that case, it's up to the serializer implementation to provide it (or else return an error).

BurntSushi (approved):

PR #632, opened 2026-08-06
test: make most tests use `-p jiff`

BurntSushi:
This was an oversight from when I switched to a virtual manifest in
cebda72f8f7702ac618e5a54ac844d0853c89de7. Specifically, because of
feature unification, these tests ended up not testing what we want:
specific feature combinations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -560,6 +560,7 @@ jobs:
     env:
       CARGO_BUILD_TARGET: wasm32-unknown-emscripten
       CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER: node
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -560,6 +560,7 @@ jobs:
     env:
       CARGO_BUILD_TARGET: wasm32-unknown-emscripten
       CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_RUNNER: node
+      CARGO_TARGET_WASM32_UNKNOWN_EMSCRIPTEN_LINKER: em++
```
BurntSushi, line 563: This fixes another problem on master that seems to have started to occur with emscripten 6.0.6. Ref https://github.com/emscripten-core/emscripten/commit/6ec69f5398fa177815194eaa81bf33da6982c601

PR #633, opened 2026-08-06
Add optional arbitrary::Arbitrary implementations

### LINKED ISSUE #631: Provide `arbitrary::Arbitrary` impls for core types
Hi, I'm using the `proptest` library for property based testing, together with the `arbitrary` crate (and the `proptest-arbitrary-interop` to make these work together) and I would like to derive the `Arbitrary` trait for my struct that hold `jiff` types. 

Would you open to a PR that optionally implemented the `Arbitrary` trait on `jiff`'s the core types? The `arbitrary` crate is fairly popular and has a stable api (1.0 released in 2021) so I don't think this will create a lot of churn.

I went ahead and checked how much work this would be. For types that are Plain Old Data it's a simple as adding a conditional derive but for `TimeZone` I'm not really sure what to do. I'm not very familiar with the inner workings of `arbitrary` so I would be fine with not implementing `Arbitrary` for it.

avsaase:
Closes https://github.com/BurntSushi/jiff/issues/631

In contrast to what I said in the issue this obviously doesn't work with simply deriving the trait since not all values of the underlying primitives are valid `jiff` types. Manually implementing the trait was easy enough though.

I had an LLM generate some unit tests of the implementations just to see what they would look like. IMO it's a bit overkill so I moved this to a separate commit that I'm happy to revert if you want/

### DIFF
--- a/crates/jiff-core/Cargo.toml
+++ b/crates/jiff-core/Cargo.toml
@@ -22,11 +22,13 @@ bench = false
 default = ["std", "tz-fat"]
 alloc = ["defmt?/alloc"]
 std = ["alloc", "log?/std"]
+arbitrary = ["dep:arbitrary"]
 defmt = ["dep:defmt"]
 logging = ["dep:log"]
 tz-fat = []
 
 [dependencies]
+arbitrary = { version = "1.4.2", optional = true, features = ["derive"] }
 defmt = { version = "1.0.0", optional = true }
 log = { version = "0.4.21", optional = true, default-features = false }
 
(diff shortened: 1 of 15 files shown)


BurntSushi:
I opened #634 because I couldn't force push to this PR for some reason.

avsaase:
Aye. Did you change anything else than add to the changelog?

avsaase:
> I opened #634 because I couldn't force push to this PR for some reason.

Perhaps that's because my fork lives under my employer's organisation and outside contributors don't have write access there.

BurntSushi:
There were a few other changes:

* Documentation in Jiff's module docs.
* Enabling `arbitrary` requires enabling `std`.
* Added a `Arbitrary` impl for `jiff::civil::Weekday`.

avsaase:
Awesome. Thanks for taking this change.

